PR #65, opened 2020-10-05
fixbug  add_row  method for loop add

wxhou:
RecursionError: maximum recursion depth exceeded in comparison


hugovk:
Thanks for the PR!

Please could you add a test that fails without the change and passes with it?

wxhou:
@hugovk  Thanks very much !  I'm not to run a test.

PR #66, opened 2020-10-06
Add setuptools as a dependency

ksunden:
`pkg_resources` is imported (to get the version number), which is provided by setuptools


ksunden:
In particular, this caused problems using `flit` which builds in an isolated environment, with no setuptools by default (and imports the package to get the version...)

hugovk:
Thank you!

hugovk:
:rocket: Released in https://pypi.org/project/prettytable/1.0.1/

PR #83, opened 2020-11-16
Support formatting options in get_json_string()

kluelesskk:
This change passes through keyword arguments to json.dumps() in a
similar manner to get_csv_string(). Additionally, it respects the
'header' table option.


hugovk:
Thank you!

Issue #86, opened 2020-12-11
Add colorful tables support


BD103:
Hi! Just as an idea, maybe you could add Colorama/Blessings/color code support for tables. Here's my idea.
```python
from prettytable import PrettyTable, THEME
x = PrettyTable()
x.set_theme(THEME.OCEAN)
x.field_names = ["Name", "Age"]
x.add_row(["Jeff", 20])
x.add_row(["Joe", 19])
x.add_row(["Jim", 25])
x.add_row(["Liam", 10])
```

This would return something like this:
![Colorful Table UWU](https://user-images.githubusercontent.com/59022059/101846410-ddbada00-3b1e-11eb-825c-8286ea2d7d94.png)

You could also have theme creation:
```python
from colorama import FORE, init
init()

class THEME:
  OCEAN = {
    "default": FORE.LIGHTCYAN_EX,
    "border": FORE.BLUE,
    "decor": FORE.CYAN
  }
```

Some possible theme class creation variables:
default = everything else
border = main border
outline = border surrounding entire table, not inner lines
hborder = horizontal border
vborder = vertical border
decor = the "+" complimenting the tables
header = the header of the table

Note: this is the program i used to render the table.
```python
y = '''
+------+-----+
| Name | Age |
+------+-----+
| Jeff |  20 |
| Joe  |  19 |
| Jim  |  25 |
| Liam |  10 |
+------+-----+
'''

blue = ["-", "|"]
cyan = ["+"]
lightcyan = ["0", "1", "2" "3", "4", "5", "6", "7", "8", "9", "2"]

z = ""

for i in y:
  if i == "\n":
    z += "\n"
  elif i in blue:
    z += "\033[34m" + i
  elif i in cyan:
    z += "\033[36m" + i
  elif i in lightcyan:
    z += "\033[96m" + i
  else:
    z += "\033[96m" + i

print(z)
```

I hope you enjoy this idea!
###### (I might fork this project later on and implement it myself!)

BD103:
This enhancement is currently waiting to be merged with the head branch. Thanks for the awesome feedback! :D

BD103:
Re-opening until PR #140 is merged

PR #87, opened 2020-12-17
Colorful Tables and Themes

BD103:
I recently suggested a new enhancement (#86) and now I've created it.

It includes:

- Themes (colorful tables)
- Easy Customization
- Kwarg included (so you can do `x = PrettyTable(theme = THEME.NAME)`)
- A new class for themes specifically
- Tutorial for creating / using / customizing / adding themes

:D


BD103:
Hm. Not sure why the tests are failing.
I don't know much about your testing system, but I did a `python setup.py install` and tested all of the features I added there. I'm hoping those tests failed because of another reason.

hugovk:
You can run the tests with:

```console
$ python -m pip install pytest pytest-cov
...
$ pytest
...
$ pytest --cov prettytable --cov tests # or with coverage
...
```


For example:
```console
$ pytest --cov prettytable --cov tests
======================================== test session starts ========================================
platform darwin -- Python 3.9.1, pytest-6.2.0, py-1.9.0, pluggy-0.13.1
Test order randomisation NOT enabled. Enable with --random-order or --random-order-bucket=<bucket_type>
rootdir: /Users/hugo/github/prettytable
plugins: requests-mock-1.8.0, random-order-1.0.4, rerunfailures-9.1.1, flaky-3.7.0, hypothesis-5.41.3, xdist-2.1.0, cov-2.10.1, testmon-1.0.3, forked-1.3.0
collected 73 items

tests/test_prettytable.py ................................................................... [ 91%]
......                                                                                        [100%]

---------- coverage: platform darwin, python 3.9.1-final-0 -----------
Name                             Stmts   Miss  Cover
----------------------------------------------------
src/prettytable/__init__.py          4      0   100%
src/prettytable/prettytable.py    1094    169    85%
tests/test_prettytable.py          423      0   100%
----------------------------------------------------
TOTAL                             1521    169    89%


======================================== 73 passed in 1.80s =========================================
```

From the CI failures, looks like colour is being added to the basic use, where no colour output is expected.

BD103:
Ok. Thanks. Though I got to say, fixing all the tests is going to be annoying. Here we go :/

BD103:
Also, pytest is failing because you moved the package into the folder `src`

hugovk:
> Also, pytest is failing because you moved the package into the folder `src`

That's intentional to make sure you're really testing the local, modified version and not a system install one.

So, from the root directory, run:

```sg
pip install -e .
```

That tells pip to install your local copy found at `.`, it looks in the `setup.py` to know the source is in `src`.

And the `-e` means "editable install", this means that when you make changes to your source, you don't need to `pip install` again, your changes will be reflected in what you test.

Then you can run `pytest` and it'll test your local changes.

src/prettytable/prettytable.py
```diff
@@ -183,9 +208,14 @@ def __init__(self, field_names=None, **kwargs):
         self._left_padding_width = kwargs["left_padding_width"] or None
         self._right_padding_width = kwargs["right_padding_width"] or None
 
-        self._vertical_char = kwargs["vertical_char"] or "|"
-        self._horizontal_char = kwargs["horizontal_char"] or "-"
-        self._junction_char = kwargs["junction_char"] or "+"
+        # color codes to theme color, then back to base
+        # _theme["border"] + kwargs["char"] + _theme["base"]
+        self._theme = kwargs["theme"] or THEME.DEFAULT
+        # Converted to try except switch because the "or" keyword wasn't working
```
hugovk: Is this comment still valid? I don't see a `try`/`except`

setup.py
```diff
@@ -26,7 +26,7 @@ def local_scheme(version):
     package_dir={"": "src"},
     use_scm_version={"local_scheme": local_scheme},
     setup_requires=["setuptools_scm"],
-    install_requires=["setuptools", "wcwidth"],
+    install_requires=["setuptools", "wcwidth", "colorama"],
```
hugovk: We'd probably want to put this as an extra (see `tests` in `extras_require` below) so it's only installed when people actually want to use, but we can deal with that later.

src/prettytable/prettytable.py
```diff
@@ -44,6 +45,12 @@
 
 import wcwidth
 
+# Converts color codes to system default (for windows)
+from colorama import init as colorinit
+
+# Changed to colorinit in case function init() is ever necessary
+colorinit()
```
hugovk: And we'd only want to call (and import?) this when actually using colours. Again, we can deal with that later.

src/prettytable/prettytable.py
```diff
@@ -183,9 +208,14 @@ def __init__(self, field_names=None, **kwargs):
...
-        self._horizontal_char = kwargs["horizontal_char"] or "-"
-        self._junction_char = kwargs["junction_char"] or "+"
+        # color codes to theme color, then back to base
+        # _theme["border"] + kwargs["char"] + _theme["base"]
+        self._theme = kwargs["theme"] or THEME.DEFAULT
+        # Converted to try except switch because the "or" keyword wasn't working
+        self._raw_vertical_char = kwargs["vertical_char"] or "|"
+        self._raw_horizontal_char = kwargs["horizontal_char"] or "-"
+        self._raw_junction_char = kwargs["junction_char"] or "+"
+        self.set_theme(kwargs["theme"] or THEME.DEFAULT)
```
hugovk: Maybe some of the test failures are coming because we're setting a default theme here, and `set_theme` then inserts some colour reset codes.

One idea: for the no-colour case, instead of setting a default theme with `""` for the colours, ditch a default theme and set `self._theme = None` to begin with.

And then only set, for example, `self._vertical_char` values inside `set_theme`, when actually called to set a theme. (Or perhaps check if a theme is set when returning a value in `vertical_char()`.)

src/prettytable/prettytable.py
```diff
@@ -1324,7 +1355,9 @@ def get_string(self, **kwargs):
                 tmp.extend(line.split("\n"))
             lines = ["|" + line[1:-1] + "|" for line in tmp]
 
-        return "\n".join(lines)
+        # added reset color code so that it doesn't overflow into other print statements
+
+        return "\n".join(lines) + "\033[0m"
```
hugovk: We don't want to return any resets when there's no theme.

hugovk (commented):
Another tip: if you just want to run one test in isolation, this runs any tests matching the string that comes after `-k`:

```sh
pytest -k testAddFieldNamesLater
```

src/prettytable/prettytable.py
BD103: No, not anymore. Let me change that.

src/prettytable/prettytable.py
BD103: Maybe we can split it into separate modules?

src/prettytable/prettytable.py
BD103: Fixed

src/prettytable/prettytable.py
BD103: Fixed

BD103:
Btw, I'm not entirely sure Black fixes the Linter issue. I've formatted the whole repo and it still gives an error.

hugovk:
> Btw, I'm not entirely sure Black fixes the Linter issue. I've formatted the whole repo and it still gives an error.

The linter does more than just Black, there's failures from isort (import sorter), Flake8 (style checker) and Prettier (like Black for non-Python code).

You can run all the lint things with:

```sh
python -m pip install -U pre-commit
pre-commit run --all-files
```

That should fix isort and Prettier, and it should be easy enough to fix the Flake8 things.

(Optionally, you can run `pre-commit install` to automatically run these things when you commit.)

setup.py
BD103: Fixing

BD103:
Ok, I think this pull request is ready

README.md
```diff
@@ -587,6 +587,64 @@ list slicing notation:
...
+
+```python
+from prettytable import PrettyTable, THEME
+```
+
+> Note, if you are using Windows, type `pip install prettytable[windows]`, so the color
+> codes get reformatted to your OS. If color does not show for other operating systems,
+> you can still do this and that may fix your problem.
+
+### Using Themes
```
```suggestion
### Using themes
```

README.md
```diff
@@ -587,6 +587,64 @@ list slicing notation:
 new_table = old_table[0:5]
 ```
 
+## Themes
+
+### Getting Themes
+
+If you want your tables to be even fancier (if that's possible) then you can use color
+themes. All you have to do is import it.
```
```suggestion
Import a theme to add color to your tables.
```

src/conftest.py
```diff
@@ -0,0 +1 @@
+# Keep empty, this tells pytest to test here
```
hugovk: Please could you show why this file is needed? What error do you get without it? I'm able to test without it.

README.md
```diff
@@ -587,6 +587,64 @@ list slicing notation:
...
+
+```python
+from prettytable import PrettyTable, THEME
+```
+
+> Note, if you are using Windows, type `pip install prettytable[windows]`, so the color
+> codes get reformatted to your OS. If color does not show for other operating systems,
+> you can still do this and that may fix your problem.
```
hugovk, line 602: If I `pip install prettytable[windows]` on non-Windows, will it make a difference?

The `init()` is only done for Windows OS.

https://github.com/tartley/colorama#initialisation says:

> Applications should initialise Colorama using ... `init()`
>
> On Windows, calling init() will filter ANSI escape sequences out of any text sent to `stdout` or `stderr`, and replace them with equivalent Win32 calls.
> 
> On other platforms, calling `init()` has no effect (unless you request other optional functionality; see "Init Keyword Args", below). By design, this permits applications to call init() unconditionally on all platforms, after which ANSI output should just work.
> 
> To stop using Colorama before your program exits, simply call `deinit()`. This will restore `stdout` and `stderr` to their original values, so that Colorama is disabled. To resume using Colorama again, call `reinit()`; it is cheaper than calling `init()` again (but does the same thing).

Some thoughts:

We could just call `init()` for all operating systems.

Or better, put it in a `try`/`except ImportError` for in case colorama isn't installed.

And in fact, we only want to call `init()` when using colour themes. If we're using plain mono, we don't want any extra import overhead.

So maybe only attempt the import/`init()` inside `set_theme`?

We could also do it when `theme != None`, and call `deinit()` when `set_theme(None)`.

(Maybe also guard the import/`init` with a `platform.system() == "Windows"`?)

setup.py
```diff
@@ -27,7 +27,7 @@ def local_scheme(version):
     use_scm_version={"local_scheme": local_scheme},
     setup_requires=["setuptools_scm"],
     install_requires=["setuptools", "wcwidth"],
-    extras_require={"tests": ["pytest", "pytest-cov"]},
+    extras_require={"tests": ["pytest", "pytest-cov"], "windows": ["colorama"]},
```
hugovk: I think we could call this `color`, because even though colorama is only needed for Windows, it's only needed for colour and not all Windows installs.

```suggestion
    extras_require={"tests": ["pytest", "pytest-cov"], "color": ["colorama"]},
```

In fact, let's make `color` also depend on Windows, something like:

https://discuss.python.org/t/can-extras-require-values-be-os-dependant/2718/2?u=hugovk

README.md
```diff
@@ -587,6 +587,64 @@ list slicing notation:
...
+
+### Using Themes
+
+After you are certain that you've imported the theme, you can quickly apply it. Now,
+printing the theme will include color.
```
```suggestion
After you've imported the theme, apply it. Now
printing the theme will include color.
```

README.md
```diff
@@ -587,6 +587,64 @@ list slicing notation:
...
+```
+
+You can also set theme on import.
+
+```python
+x = PrettyTable(theme = THEME.OCEAN)
+print(x)
+```
+
+### Custom Themes
```
```suggestion
### Custom themes
```

README.md
```diff
@@ -587,6 +587,64 @@ list slicing notation:
...
+my_theme = {
+  "base": "\033[31m",
+  "border": "\033[33m",
+  "decor": "\033[0m"
+}
+x.set_theme(my_theme)
+print(x)
+```
+
+This would print something like the following:
```
```suggestion
This prints something like this:
```

README.md
```diff
@@ -596,3 +654,6 @@ changed lines.
 python -m pip install black
 black prettytable*.py
 ```
+
+Adding themes are pretty easy. Just add a dictionary just like in the custom theme
+tutorial under the `THEME` class. :D
```
hugovk: Let's skip this bit :)

```suggestion
```

README.md
```diff
@@ -587,6 +587,64 @@ list slicing notation:
...
+print(x)
+```
+
+This would print something like the following:
+
+![Theme Sample](https://user-images.githubusercontent.com/59022059/102506142-d0d34480-4050-11eb-8b2e-4a5ee7ab391f.png)
+
+Replace the sample color codes with your own colors.
+
+### List of Themes
```
```suggestion
### List of themes
```

src/prettytable/prettytable.py
```diff
@@ -68,6 +76,26 @@ def _get_size(text):
     return width, height
 
 
+class THEME:
+    #   Old Default
+    #   DEFAULT = {"base": "", "border": "", "decor": ""}
+
```
hugovk: Delete if not needed:
```suggestion
```

README.md
```diff
@@ -587,6 +587,64 @@ list slicing notation:
...
+
+```python
+x = PrettyTable(theme = THEME.OCEAN)
+print(x)
+```
+
+### Custom Themes
+
+If you don't want any of the themes supplied, you can make your own. Create a custom
+dictionary with the following keys:
```
hugovk, line 624: Maybe mention what the keys are with a link to a reference guide?

README.md
```diff
@@ -587,6 +587,64 @@ list slicing notation:
...
+
+```python
+my_theme = {
+  "base": "\033[31m",
+  "border": "\033[33m",
+  "decor": "\033[0m"
```
hugovk: The code uses "junction" for "decor", and it's also exposed via the API, so let's replace all "decor" with junction".

```suggestion
  "junction": "\033[0m"
```

hugovk (changes requested):
Thanks for the update, I think we're very close now! Here's a few more suggestions.

src/conftest.py
BD103: Oh, I added this so that pytest knew where to test. I must have had a weird version, or something. Let me see if it's necessary,

README.md
```diff
@@ -628,7 +628,7 @@ dictionary with the following keys:
 my_theme = {
   "base": "\033[31m",
   "border": "\033[33m",
-  "decor": "\033[0m"
+  "junction": "\033[0m"
```
hugovk, line 630: Please also rename `decor` -> `junction` in the code.

README.md
```diff
@@ -587,6 +587,63 @@ list slicing notation:
 new_table = old_table[0:5]
 ```
 
+## Themes
+
+### Getting Themes
```
```suggestion
### Importing themes
```

setup.py
```diff
@@ -27,7 +27,7 @@ def local_scheme(version):
     use_scm_version={"local_scheme": local_scheme},
     setup_requires=["setuptools_scm"],
     install_requires=["setuptools", "wcwidth"],
-    extras_require={"tests": ["pytest", "pytest-cov"]},
+    extras_require={"tests": ["pytest", "pytest-cov"], "color": ["colorama"]},
```
hugovk: Let's make it also depend on Windows, something like:

https://discuss.python.org/t/can-extras-require-values-be-os-dependant/2718/2?u=hugovk

```suggestion
    extras_require={"tests": ["pytest", "pytest-cov"], "color": ["colorama; sys_platform == "win32"]},
```

README.md
BD103: What kind of reference? For color codes, custom themes, or what?

BD103:
Btw, I will not be as active for winter break, so some of these changes will take time. 😄

src/conftest.py
BD103: Removed

README.md
BD103: I found a way to implement this finally.
I found this solution in a separate [project](https://github.com/Team-Engage/engine/blob/3de2e93d769252539ab6cabe4f515980e5bc2915/setup.py#L19), and tested it on windows. Should work now, though I must admit the color codes do not translate as smoothly as I was hoping. That should be find though.

README.md
BD103: Just as a follow up, this is the last thing that you wanted me to do before I format and finish. Can you please explain what you mean by this?
Thanks!

BD103:
@hugovk can you review?

hugovk:
@BD103 Yes, hopefully I can review this weekend, thank you for your patience!

README.md
hugovk: Yeah, a link to some page that lists the different sort of values for `"\033[31m"`, `"\033[33m"`, etc. so if I want to make my own theme I know what codes to use for yellow, red, etc.

src/prettytable/prettytable.py
```diff
@@ -1768,6 +1794,50 @@ def _get_formatted_html_string(self, options):
...
+        self._theme = theme
+        try:
+            self._vertical_char = (
+                self._theme["border"]
+                + self._raw_vertical_char
+                + "\033[0m"
+                + self._theme["base"]
+            )
+        except TypeError:
+            self._vertical_char = (
```
hugovk, line 1815: Code coverage indicates this except block is not covered by tests.

Under what circumstances could it occur?

I guess if any `self._raw_vertical_char` isn't a string?

How likely is this, and how could it occur? Is it okay to just let the `TypeError` raise up?

Is it for when a custom theme doesn't define this particular character?

(Similarly for the other two `except` blocks.)

src/prettytable/prettytable.py
```diff
@@ -1768,6 +1794,50 @@ def _get_formatted_html_string(self, options):
...
         return "\n".join(lines)
 
+    def set_theme(self, theme):
+        # Escapes if there is no theme
+        if theme is None:
+            self._theme = None
+            self._vertical_char = self._raw_vertical_char
+            self._horizontal_char = self._raw_horizontal_char
+            self.junction_char = self._raw_junction_char
+            return ""
```
hugovk: This doesn't need to return a value
```suggestion
            return
```

tests/test_prettytable.py
```diff
@@ -26,6 +28,15 @@
...
 
+def escape_ansi(line):
+    ansi_escape = re.compile(r"(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]")
+    return ansi_escape.sub("", line)
+
+
+def test_escape_ansi():
+    assert escape_ansi("\033[31mhithere uwu\033[0m") == "hithere uwu"
+
+
```
hugovk: Delete this, it's not testing any part of the library
```suggestion
```

src/prettytable/prettytable.py
```diff
@@ -1768,6 +1794,50 @@ def _get_formatted_html_string(self, options):
 
         return "\n".join(lines)
 
+    def set_theme(self, theme):
+        # Escapes if there is no theme
+        if theme is None:
+            self._theme = None
+            self._vertical_char = self._raw_vertical_char
+            self._horizontal_char = self._raw_horizontal_char
+            self.junction_char = self._raw_junction_char
```
hugovk, line 1803: I think we can remove this, and rename `self._raw_vertical_char` to `self._vertical_char` (and for horizontal and junction) everywhere else?

```suggestion
```

hugovk (changes requested):

BD103:
Closing stale branch

PR #95, opened 2021-01-20
Docs: Remove reference to Python 2.x print statement in README

johnthagen:
As this project only supports Python 3.6+, remove reference to Python 2.x `print` statement to make README simpler.


hugovk:
Thanks!

PR #106, opened 2021-05-04
Add auto index column

### LINKED ISSUE #90: Suggest: auto index column
It would be good if there is an option to switch on an auto index column. (1, 2, 3 ...)

erictzimas:
Fixes #90.


erictzimas:
Should i add a test for this function? how should i handle those 5 failing checks?

hugovk:
Thanks for the PR!

Please could you add tests to `tests/test_prettytable.py`? Just let me know if you're not sure how to do that, and I can help out.

We can ignore the 3.10-dev failures for now, it's unrelated and caused by pytest not yet supporting 3.10 beta.

erictzimas:
I will work on it, thank you!

src/prettytable/prettytable.py
```diff
@@ -1087,6 +1087,30 @@ def add_column(self, fieldname, column, align="c", valign="t"):
...
+        fieldname = "Index"
+        self._validate_align(align)
+        self._validate_valign(valign)
+        self._field_names.append(fieldname)
+        self._align[fieldname] = align
+        self._valign[fieldname] = valign
+        c = 1
+        for i in range(0, len(self._rows)):
+            if len(self._rows) < i + 1:
+                self._rows.append([])
```
hugovk: Please could you add another test case (or adjust the existing one) to also cover this line?

tests/test_prettytable.py
```diff
@@ -6,6 +6,7 @@
 import unittest
 from math import e, pi, sqrt
 
+
 import pytest
```
hugovk: Fix lint:

```suggestion
import pytest
```

hugovk (changes requested):

src/prettytable/prettytable.py
erictzimas: I certainly can, but after looking into it, i think that this if statement is not needed, should I remove it? And if I do, are there any more test cases I will need to add? Thank you

hugovk: If it's not needed please remove.

Would it be useful to test different values of `align` and `valign`?

erictzimas:
I added a test case for align but I am not sure i can do it for valign since i do not fully understand its functionality.

src/prettytable/prettytable.py
erictzimas: I added a test case for align but I am not sure i can do it for valign since i do not fully understand its functionality.

hugovk: `valign` is vertical alignment for the table. For example, top alignment may looks something a bit like this (a and b are in the same cell, 2 is another cell, 3 a third):

```
a   2   3   
b
```

Bottom alignment something like this:

```
a
b   2   3   
```

But come to think about it, why are we modifying properties of the table when adding an autoindex column? Can we remove these parameters, or do you think they're necessary?

hugovk:
So this takes a table like this:

```
+-----------+------+------------+-----------------+
| City name | Area | Population | Annual Rainfall |
+-----------+------+------------+-----------------+
|  Adelaide | 1295 |  1158259   |      600.5      |
|  Brisbane | 5905 |  1857594   |      1146.4     |
|   Darwin  | 112  |   120900   |      1714.7     |
|   Hobart  | 1357 |   205556   |      619.5      |
|   Sydney  | 2058 |  4336374   |      1214.8     |
| Melbourne | 1566 |  3806092   |      646.9      |
|   Perth   | 5386 |  1554769   |      869.4      |
+-----------+------+------------+-----------------+
```

And adds an index to the right column:

```
+-----------+------+------------+-----------------+-------+
| City name | Area | Population | Annual Rainfall | Index |
+-----------+------+------------+-----------------+-------+
|  Adelaide | 1295 |  1158259   |      600.5      |   1   |
|  Brisbane | 5905 |  1857594   |      1146.4     |   2   |
|   Darwin  | 112  |   120900   |      1714.7     |   3   |
|   Hobart  | 1357 |   205556   |      619.5      |   4   |
|   Sydney  | 2058 |  4336374   |      1214.8     |   5   |
| Melbourne | 1566 |  3806092   |      646.9      |   6   |
|   Perth   | 5386 |  1554769   |      869.4      |   7   |
+-----------+------+------------+-----------------+-------+
```

However, @yanhuihang's example in https://github.com/jazzband/prettytable/issues/90 showed the index on the left column:

```
+-------+----------------------+----------------------+
| Index |          a           |          b           |
+-------+----------------------+----------------------+
|   1   |  0.8373683108488006  | -1.9662674211289342  |
|   2   |  0.8166690183400099  | -0.6827019085998253  |
|   3   | -0.34864565952912674 | -0.3828836864526866  |
|   4   |  -0.786195405989435  | -0.28278650269529015 |
|   5   | -0.7737543534146584  | 0.39198996847252016  |
|   6   |  1.2756295710689314  | 0.24976436160178775  |
|   7   |  -2.24766869070198   | -1.0958173510686795  |
|   8   | -0.2123610818363809  | -0.17583602522710734 |
|   9   | -0.9998924564529872  |  -0.78502826234026   |
|   10  |  -0.850991227375058  |  0.3267603399304381  |
+-------+----------------------+----------------------+
```

Would it be better on the left?

And should it be possible to set the column's field name to something else via a parameter? It can default to "Index".

erictzimas:
Taking everything you said into consideration, I think i will remove valign and align as it is not necessary to have a separate formatting in the index column, i will add a parameter that represents the name of the index column and i will try to add the column on the left as I agree, this how it is normally done. Thank you very much for the guidance.

erictzimas:
I modified the function to add the column on the left, and added the `fieldname` parameter, i did not remove `valign` and `align` since it does not seem to work if i do not specify this attributes when adding a new column as in `add_column()`

```python
def add_column(self, fieldname, column, align="c", valign="t"):

        """Add a column to the table.

        Arguments:

        fieldname - name of the field to contain the new column of data
        column - column of data, should be a list with as many elements as the
        table has rows
        align - desired alignment for this column - "l" for left, "c" for centre and
            "r" for right
        valign - desired vertical alignment for new columns - "t" for top,
            "m" for middle and "b" for bottom"""

        if len(self._rows) in (0, len(column)):
            self._validate_align(align)
            self._validate_valign(valign)
            self._field_names.append(fieldname)
            self._align[fieldname] = align
            self._valign[fieldname] = valign
            for i in range(0, len(column)):
                if len(self._rows) < i + 1:
                    self._rows.append([])
                self._rows[i].append(column[i])
        else:
            raise Exception(
                f"Column length {len(column)} does not match number of rows "
                f"{len(self._rows)}"
            )
```

If that is ok please inform me so i can create a PR that also covers the `_valign` parameter, thank you.

src/prettytable/prettytable.py
```diff
@@ -1087,6 +1087,27 @@ def add_column(self, fieldname, column, align="c", valign="t"):
...
+
+        self._validate_align(align)
+        self._validate_valign(valign)
+        self._field_names.insert(0, fieldname)
+        self._align[fieldname] = align
+        self._valign[fieldname] = valign
+        c = 1
+        for i in range(0, len(self._rows)):
+            self._rows[i].insert(0, c)
+            c += 1
```
hugovk: How about using the existing values of `self.align` and `self.valign`?

For example:

```suggestion
    def add_autoindex(self, fieldname="Index"):

        """Add an auto-incrementing index column to the table.

        Arguments:

        fieldname - name of the field to contain the new column of data"""

        self._field_names.insert(0, fieldname)
        self._align[fieldname] = self.align
        self._valign[fieldname] = self.valign
        for i, row in enumerate(self._rows):
            row.insert(0, i + 1)
```

This also simplifies the loop a bit too.

The tests will need updating too.

erictzimas: I installed the latest development version and it tracked some other changes that were made in the meantime, apart from mine, I'm sorry for the inconvenience. Do you think we should also add testing for different valign and align values of the table or is it not needed?

src/prettytable/prettytable.py
```diff
@@ -337,39 +337,39 @@ def _validate_field_names(self, val):
         try:
             assert len(val) == len(set(val))
         except AssertionError:
-            raise Exception("Field names must be unique")
+            raise Exception("Field names must be unique!")
```
hugovk: Please don't add exclamation marks to exception here and elsewhere, they were only recently removed in https://github.com/jazzband/prettytable/pull/101.

hugovk (changes requested):

src/prettytable/prettytable.py
hugovk: I don't think we need to test different valign and align, should be enough they're tested elsewhere. Thanks!

src/prettytable/prettytable.py
erictzimas: This was the version problem i was talking about, I reverted the changes.

src/prettytable/prettytable.py
erictzimas: Thank you!

hugovk:
Thank you very much!

mailsanchu:
Is this index broken if you apply sort?

hugovk:
It adds a column to the data, so if you sort after adding, yes. So it's best sort first and add the index last.

mailsanchu:
Cool Is it possible to formatted index like 001, 002 etc ?

hugovk:
No.

mailsanchu:
No worries

hugovk:
You can always manually add a column with whatever data you want, including numbers formatted however you like.

mailsanchu:
Yes I have done exactly that. Thanks 

```python
my_list.sort(key=lambda a: a[1])
for idx, val in enumerate(my_list):
    val.insert(0, f"{(idx + 1):0>3}")
```

Issue #107, opened 2021-05-05
Add function that adds column of average values in table


erictzimas:
Hello, i would suggest creating a custom function that adds a column of the averages of each row to the table if the values are numerical. I will create a PR to showcase such functionality.

hugovk:
Hi, thanks for the suggestion!

I think I might decline this as it seems pretty specialised and PrettyTable isn't intended to be a spreadsheet. And it should be straightforward to pre-calculate whatever values are needed and add them as a column.

PR #109, opened 2021-05-05
Add function that adds column with the average of each row as per #107

erictzimas:
Add a new function that adds a column with the average of each row


hugovk:
Closing per https://github.com/jazzband/prettytable/issues/107#issuecomment-832692515. Thanks anyway!

PR #104, opened 2021-04-27
Add support for positional junction characters and add "double_border" table style

guysz:
The positional junction characters provides the user more control of tables' layout in the terminal.
The "double_border" style is a usage example of the positional junction characters.

PR #115, opened 2021-05-31
Access properties instead of protected fields in _get_options()

guysz:
Required by PR #104

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -449,6 +449,16 @@ class PrettyTable:
     ##############################
 
     @property
+    def xhtml(self):
+        """Print <br/> tags if True, <br> tags if false"""
+        return self._xhtml
+
+    @xhtml.setter
+    def xhtml(self, val):
+        self._validate_option("xhtml", val)
+        self._xhtml = val
+
+    @property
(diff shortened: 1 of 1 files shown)


src/prettytable/prettytable.py
```diff
@@ -448,6 +448,16 @@ def _validate_attributes(self, name, val):
     # ATTRIBUTE MANAGEMENT       #
     ##############################
 
+    @property
+    def xhtml(self):
+        """Print <br/> tags if True, <br> tags if false"""
```
```suggestion
        """Print <br/> tags if True, <br> tags if False"""
```
? And in the other place this text is used?

hugovk (commented):
This adds a new setter/getter to the public API.

In addition to being useful for #104, what is the motivation? Is it for consistency with other setters/getters of similar properties?

src/prettytable/prettytable.py
guysz: Fixed the docstring.

Regarding your first question - 
Yes this change makes the code more consistent with its design. The "xhtml" option is the only option that is not exposed as a property, and I believe it could be useful.
This consistency allows implementing #104 in a more elegant way.

hugovk:
Okay, thanks!

PR #117, opened 2021-06-06
Use declarative metadata and support PyPy3

hugovk:
Plus update some config


hugovk:
Coverage went down, but that's because lines of (covered) Python code went down by moving them from `setup.py` to `setup.cfg` which is intended.

Issue #116, opened 2021-06-05
Set the width of each column


TimFang4162:
I really need this function.

scastria:
Isn't that what `max_width` and `min_width` are for which can be set on a per column basis with a dictionary?

hugovk:
Closing, the issue doesn't have enough info, and if `max_width`/`min_width` don't help, please provide more info.

@scastria Thanks!

PR #126, opened 2021-09-13
Set the length of each column

zzzkorn:
Set the length of each column with a separator `'`. For example:
```python
x = PrettyTable()
x.field_names = ["Country'50", "Capital", "is_russia", "that's mine"]
x.add_row(["Russia", "Moscow", True, True])
x.add_rows([["Argentina", "Buenos Aires", False, False], ["Jamaica", "Kingston", False, False]])
x.add_column("Starts with A'77", [False, True, False])
x.add_column("Can't do this'30", [False, True, False])

print(x)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,12 +354,12 @@ relevant row, in the column specified by the `sort_by` argument. The remaining n
 elements are the data in each of the table's columns, in order, including a repeated
 instance of the data in the `sort_by` column.
 
-##### Setting the minimum columns lengths
+##### Setting columns' minimum widths
 
 You can set the minimum width of each column using the property `columns_min_widths`:
 
 ```python
-x.columns_min_widths = [30, 15, 20, None]
+x.columns_min_widths = [30, 15, 20, 0]
 print(x)
 ```
 
@@ -379,7 +379,8 @@ to get:
 +--------------------------------+-----------------+----------------------+-----------------+
 ```
 
-If the argument `None` is passed, then the minimum width is automatically selected
+If `0` is given for a column, then the minimum width is calculated based on the column's
+data.
 
 ### Changing the appearance of your table - the easy way
 
--- a/README.md
+++ b/README.md
@@ -355,7 +355,9 @@ elements are the data in each of the table's columns, in order, including a repe
 instance of the data in the `sort_by` column.
 
 ##### Setting the minimum columns lengths
+
 You can set the minimum width of each column using the property `columns_min_widths`:
+
 ```python
 x.columns_min_widths = [30, 15, 20, None]
 print(x)
@@ -376,6 +378,7 @@ to get:
 |             Perth              |       5386      |       1554769        |      869.4      |
 +--------------------------------+-----------------+----------------------+-----------------+
 ```
+
 If the argument `None` is passed, then the minimum width is automatically selected
 
 ### Changing the appearance of your table - the easy way
--- a/README.md
+++ b/README.md
@@ -354,6 +354,30 @@ relevant row, in the column specified by the `sort_by` argument. The remaining n
 elements are the data in each of the table's columns, in order, including a repeated
 instance of the data in the `sort_by` column.
 
+##### Setting the minimum columns lengths
+You can set the minimum width of each column using the property `columns_min_widths`:
+```python
+x.columns_min_widths = [30, 15, 20, None]
+print(x)
+```
+
+to get:
+
+```
++--------------------------------+-----------------+----------------------+-----------------+
+|           City name            |       Area      |      Population      | Annual Rainfall |
++--------------------------------+-----------------+----------------------+-----------------+
+|            Adelaide            |       1295      |       1158259        |      600.5      |
+|            Brisbane            |       5905      |       1857594        |      1146.4     |
+|             Darwin             |       112       |        120900        |      1714.7     |
+|             Hobart             |       1357      |        205556        |      619.5      |
+|             Sydney             |       2058      |       4336374        |      1214.8     |
+|           Melbourne            |       1566      |       3806092        |      646.9      |
+|             Perth              |       5386      |       1554769        |      869.4      |
++--------------------------------+-----------------+----------------------+-----------------+
+```
+If the argument `None` is passed, then the minimum width is automatically selected
+
 ### Changing the appearance of your table - the easy way
 
 By default, PrettyTable produces ASCII tables that look like the ones used in SQL
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -126,6 +126,7 @@ class PrettyTable:
         self.encoding = kwargs.get("encoding", "UTF-8")
 
         # Data
+        self._user_column_widths = []
         self._field_names = []
         self._rows = []
         self.align = {}
@@ -507,6 +508,30 @@ class PrettyTable:
         except AssertionError:
             raise Exception("Attributes must be a dictionary of name/value pairs")
 
+    def _get_column_name_wight(self, field):
+        width, name = None, str()
+        split_field = field.split("'")
+        if len(split_field) > 1:
+            try:
+                width = int(split_field[-1])
+                name = "'".join(split_field[0:-1])
+            except ValueError:
+                name = field
+        else:
+            name = field
+        return name, width
+
+    def _get_column_name_and_set_wight(self, val):
+        if isinstance(val, str):
+            temp_val, temp_width = self._get_column_name_wight(val)
+            self._user_column_widths.append(temp_width)
+        else:
+            temp_val = []
+            for v in val:
+                temp_name, temp_width = self._get_column_name_wight(v)
+                temp_val.append(temp_name)
+                self._user_column_widths.append(temp_width)
+        return temp_val
     ##############################
     # ATTRIBUTE MANAGEMENT       #
     ##############################
@@ -532,7 +557,7 @@ class PrettyTable:
 
     @field_names.setter
     def field_names(self, val):
-        val = [str(x) for x in val]
+        val = self._get_column_name_and_set_wight([str(x) for x in val])
         self._validate_option("field_names", val)
         old_names = None
         if self._field_names:
@@ -1299,6 +1324,7 @@ class PrettyTable:
         if len(self._rows) in (0, len(column)):
             self._validate_align(align)
             self._validate_valign(valign)
+            fieldname = self._get_column_name_and_set_wight(fieldname)
             self._field_names.append(fieldname)
             self._align[fieldname] = align
             self._valign[fieldname] = valign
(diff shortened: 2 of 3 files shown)


hugovk:
Thanks for the contribution, but I think this could break existing data that just happens to have an apostrophe before a number.

zzzkorn:
Only the number after the `'` at the end of the line is taken into account. We can also use a less common separator.

For example:
```
x = PrettyTable()
x.field_names = ["Country'500 '30", "Capital", "is_russia", "that's mine"]

+--------------------------------+---------+-----------+-------------+
|          Country'500           | Capital | is_russia | that's mine |
+--------------------------------+---------+-----------+-------------+
+--------------------------------+---------+-----------+-------------+
```

hugovk:
We have no idea what characters people are using in their data.

Would you like to propose an alternative API that doesn't involve interpreting user data?

How about an API like `x.set_widths([50, None, ....])`?

And/or add a `width` parameter to `add_column` and/or `add_rows`?

How would it handle data that is longer than the requested width? Cropping the data? Or is it a minimum width?

Changes would require unit tests to cover different use cases, plus documentation in README and docstrings.

(See also https://github.com/jazzband/prettytable/issues/116.)

zzzkorn:
Ok, I'll take care of the `x.set_widths()` method implementation and related tasks.

README.md
```diff
@@ -354,6 +354,33 @@ relevant row, in the column specified by the `sort_by` argument. The remaining n
...
+|            Brisbane            |       5905      |       1857594        |      1146.4     |
+|             Darwin             |       112       |        120900        |      1714.7     |
+|             Hobart             |       1357      |        205556        |      619.5      |
+|             Sydney             |       2058      |       4336374        |      1214.8     |
+|           Melbourne            |       1566      |       3806092        |      646.9      |
+|             Perth              |       5386      |       1554769        |      869.4      |
++--------------------------------+-----------------+----------------------+-----------------+
+```
+
+If the argument `None` is passed, then the minimum width is automatically selected
```
```suggestion
If `None` is given for a column, then the minimum width is calculated based on the column's data.
```

README.md
```diff
@@ -354,6 +354,33 @@ relevant row, in the column specified by the `sort_by` argument. The remaining n
 elements are the data in each of the table's columns, in order, including a repeated
 instance of the data in the `sort_by` column.
 
+##### Setting the minimum columns lengths
```
```suggestion
##### Setting columns' minimum widths
```

src/prettytable/prettytable.py
```diff
@@ -377,6 +378,23 @@ def _validate_option(self, option, val):
...
+        if self._field_names or self._rows:
+            try:
+                assert len(val) == len(self._field_names)
+                assert len(val) == len(self._rows[0])
+            except AssertionError:
+                field_len = (
+                    len(self._field_names) if self._field_names else len(self._rows[0])
+                )
+                raise Exception(
+                    "Columns minimum widths list has incorrect number of values, "
```
```suggestion
                    "Columns' minimum widths list has incorrect number of values, "
```

README.md
hugovk: Also, we're internally translating `None` into `0`, shall we change the API to take `0` instead of `None`?

I think zero has more meaning than `None`: "I want the minimum width to be zero = I'm not setting a bigger value, it has to be *at least zero*!"

src/prettytable/prettytable.py
```diff
@@ -511,6 +529,16 @@ def _validate_attributes(self, name, val):
     # ATTRIBUTE MANAGEMENT       #
     ##############################
 
+    @property
+    def columns_min_widths(self):
+        return self._columns_min_widths
+
+    @columns_min_widths.setter
+    def columns_min_widths(self, val):
+        val = [int(x) if x else 0 for x in val]
```
hugovk: And if we use `0` instead of `None`, we could do something like this and allow the normal `TypeError` for `int(None)`: 
```suggestion
        val = [int(x) for x in val]
```

src/prettytable/prettytable.py
```diff
@@ -1603,10 +1634,13 @@ def _stringify_hrule(self, options, where=""):
         if not self._field_names:
             bits.append(options[where + "right_junction_char"])
             return "".join(bits)
-        for field, width in zip(self._field_names, self._widths):
+        for field, width, min_widths in zip(
```
```suggestion
        for field, width, min_width in zip(
```

src/prettytable/prettytable.py
```diff
@@ -1603,10 +1634,13 @@ def _stringify_hrule(self, options, where=""):
...
+            self._field_names, self._widths, self._columns_min_widths
+        ):
             if options["fields"] and field not in options["fields"]:
                 continue
-            bits.append((width + lpad + rpad) * options["horizontal_char"])
+            temp_width = width if width > min_widths else min_widths
```
```suggestion
            temp_width = max(width, min_width)
```

src/prettytable/prettytable.py
```diff
@@ -1662,7 +1696,11 @@ def _stringify_header(self, options):
                 bits.append(options["vertical_char"])
             else:
                 bits.append(" ")
-        for (field, width) in zip(self._field_names, self._widths):
+        for (field, width, min_widths) in zip(
```
```suggestion
        for (field, width, min_width) in zip(
```

src/prettytable/prettytable.py
```diff
@@ -1675,9 +1713,10 @@ def _stringify_header(self, options):
                 fieldname = field.lower()
             else:
                 fieldname = field
+            temp_width = width if width > min_widths else min_widths
```
```suggestion
            temp_width = max(width, min_width)
```

src/prettytable/prettytable.py
```diff
@@ -1727,8 +1766,12 @@ def _stringify_row(self, row, options, hrule):
                 else:
                     bits[y].append(" ")
 
-        for (field, value, width) in zip(self._field_names, row, self._widths):
-
+        for (field, value, width, min_widths) in zip(
```
```suggestion
        for (field, value, width, min_width) in zip(
```

src/prettytable/prettytable.py
```diff
@@ -1748,10 +1791,10 @@ def _stringify_row(self, row, options, hrule):
             for line in lines:
                 if options["fields"] and field not in options["fields"]:
                     continue
-
+                temp_width = width if width > min_widths else min_widths
```
```suggestion
                temp_width = max(width, min_widths)
```

tests/test_prettytable.py
```diff
@@ -1196,3 +1196,101 @@ def testBordered(self):
...
+                result="""
++-----------+------+------------+
+| City name | Area | Population |
++-----------+------+------------+
+|  Adelaide | 1295 |  1158259   |
+|  Brisbane | 5905 |  1857594   |
+|   Darwin  | 112  |   120900   |
++-----------+------+------------+
+""",
+                min_widths=[None, None, None],
```
```suggestion
                min_widths=[0, 0, 0],
```

tests/test_prettytable.py
```diff
@@ -1196,3 +1196,101 @@ def testBordered(self):
...
+                result="""
++----------------------+--------------------------------+------------+
+|      City name       |              Area              | Population |
++----------------------+--------------------------------+------------+
+|       Adelaide       |              1295              |  1158259   |
+|       Brisbane       |              5905              |  1857594   |
+|        Darwin        |              112               |   120900   |
++----------------------+--------------------------------+------------+
+""",
+                min_widths=[20, 30, None],
```
```suggestion
                min_widths=[20, 30, 0],
```

tests/test_prettytable.py
```diff
@@ -1196,3 +1196,101 @@ def testBordered(self):
...
+|   Darwin  |              112               |   120900   |
++-----------+--------------------------------+------------+
+""",
+                min_widths=[2, 30, 3],
+            ),
+        ]
+        for table in tables:
+            self.x.columns_min_widths = table["min_widths"]
+            result = self.x.get_string()
+            self.assertEqual(result.strip(), table["result"].strip())
```
hugovk: Let's stick with pytest style asserts:
```suggestion
            assert result.strip() == table["result"].strip()
```

tests/test_prettytable.py
```diff
@@ -1196,3 +1196,101 @@ def testBordered(self):
...
+    def testExceptionsWithFieldNames(self):
+        with self.assertRaises(Exception):
+            self.x.columns_min_widths = [20, 30]
+        with self.assertRaises(Exception):
+            self.x.columns_min_widths = [20, 30, 40, 50]
+
+    def testExceptionsWithoutFieldNames(self):
+        with self.assertRaises(Exception):
+            self.y.columns_min_widths = [20, 30]
+        with self.assertRaises(Exception):
```
```suggestion
        with pytest.raises(Exception):
            self.x.columns_min_widths = [20, 30]
        with pytest.raises(Exception):
            self.x.columns_min_widths = [20, 30, 40, 50]

    def testExceptionsWithoutFieldNames(self):
        with pytest.raises(Exception):
            self.y.columns_min_widths = [20, 30]
        with pytest.raises(Exception):
```

tests/test_prettytable.py
```diff
@@ -1196,3 +1196,101 @@ def testBordered(self):
...
+            self.y.columns_min_widths = [20, 30, 40, 50]
+
+    def testAutoIndex(self):
+        self.auto_index_table.add_autoindex(fieldname="Test")
+        self.assertEqual(self.auto_index_table.columns_min_widths, [0, 0, 0, 0])
```
```suggestion
        assert self.auto_index_table.columns_min_widths == [0, 0, 0, 0]
```

tests/test_prettytable.py
```diff
@@ -1196,3 +1196,101 @@ def testBordered(self):
...
+
+        self.auto_index_table.add_autoindex(fieldname="Test", min_width=30)
+        self.assertEqual(self.auto_index_table.columns_min_widths, [30, 0, 0, 0, 0])
```
```suggestion
        assert self.auto_index_table.columns_min_widths == [30, 0, 0, 0, 0]
```

tests/test_prettytable.py
```diff
@@ -1196,3 +1196,101 @@ def testBordered(self):
...
+
+    def testAddColumn(self):
+        self.add_column_table.add_column("Starts with A", [True, False, False])
+        self.assertEqual(self.add_column_table.columns_min_widths, [0, 0, 0, 0])
+
+        self.add_column_table.add_column(
+            "Starts with B", [False, True, False], min_width=30
+        )
+        self.assertEqual(self.add_column_table.columns_min_widths, [0, 0, 0, 0, 30])
```
```suggestion
        assert self.add_column_table.columns_min_widths == [0, 0, 0, 0]

        self.add_column_table.add_column(
            "Starts with B", [False, True, False], min_width=30
        )
        assert self.add_column_table.columns_min_widths == [0, 0, 0, 0, 30]
```

hugovk (changes requested):
Thanks! A few suggestions.

hugovk:
How does this work with `min_width` and `max_width` for the whole table?

They're documented in the docstring, but not the README: https://github.com/jazzband/prettytable/pull/98/files

I'd guess these column min widths should override the table one? Some tests mixing would be useful.

zzzkorn:
Ok i will test this

zzzkorn:
The _compute_widths() method subtracts the column width, taking into account the minimum and maximum widths. This width is started with the columns_min_width and the maximum value is taken.

PR #132, opened 2021-09-25
Consider US G0 Character Set ANSI-escape code as 0-width (like colors)

### LINKED ISSUE #129: Some ansi-escape sequences break width
### What did you do?

`blessed` is a popular python library for dealing with terminal escape codes. I use it for coloring.
However prettytable doesn't interpret some codes correctly.

The color codes are interpreted fine by prettytable as 0-width and so should the color-reset code.

### What actually happened?

The color codes are interpreted fine by prettytable as 0-width but not the color-reset code which thus breaks the width as the terminal renders 0-width but prettytable accounted multiple characters width.

* Python: 3.9
* PrettyTable: prettytable==2.2.0 (and same problem though different rendering with debian's prettytable==0.7.2)

```python
import prettytable

tab = prettytable.PrettyTable(["foo", "bar"])
tab.add_row([1, 2])
tab.add_row([3, 4])
print(tab)  # everything fine so far

import blessed
t = blessed.Terminal()
tab.add_row([f"{t.bold_red}{t.bold_red}{t.bold_red}{t.bold_red}1", "2"])
print(tab)  # width is ok, color is not finished but that's my fault, no bug here

tab.add_row([f"{t.normal}{t.normal}{t.normal}{t.normal}1", "2"])
print(tab)  # now cell width is broken!
```

hydrargyrum:
The ANSI escape code "^[(B" is used in terminal for changing the
character set, and is output by some terminal-coloring libraries as part
of color-resetting code (for example, by blessed). It's also present in
"sgr0" code of terminfo from some terminals.

The bug was that prettytable didn't interpret it as 0-width
non-printable sequence but as the sequence of individual characters,
breaking the width of the column.
Like colors codes, it should be ignored when computing the width of a
cell.

Closes #129.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -59,7 +59,7 @@ ORGMODE = 14
 DOUBLE_BORDER = 15
 RANDOM = 20
 
-_re = re.compile(r"\033\[[0-9;]*m")
+_re = re.compile(r"\033\[[0-9;]*m|\033\(B")
 
 
 def _get_size(text):
--- a/tests/test_prettytable.py
+++ b/tests/test_prettytable.py
@@ -580,6 +580,44 @@ class BreakLineTests(unittest.TestCase):
         )
 
 
+class AnsiWidthTest(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


hugovk:
Thank you! Will release this right away.

PR #135, opened 2021-10-07
Move test to pytest

davzucky:
This is a refactoring of all the test to use pytest and fixture. Just wanted to clean the test before implementing the custom format

### DIFF
--- a/tests/test_prettytable.py
+++ b/tests/test_prettytable.py
@@ -27,6 +27,8 @@ from prettytable import (
 )
 from prettytable.prettytable import FRAME
 
+import prettytable
+
 
 def helper_table(rows=3):
     t = PrettyTable(["Field 1", "Field 2", "Field 3"])
@@ -113,8 +115,8 @@ class TestBuildEquivalence():
         assert row_prettytable.get_html_string() == mix_prettytable.get_html_string()
 
 
-class DeleteColumnTest(unittest.TestCase):
-    def testDeleteColumn(self):
+class TestDeleteColumn():
+    def test_DeleteColumn(self):
         with_del = PrettyTable()
         with_del.add_column("City name", ["Adelaide", "Brisbane", "Darwin"])
         with_del.add_column("Area", [1295, 5905, 112])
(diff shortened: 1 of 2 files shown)


tests/test_prettytable.py
```diff
@@ -125,293 +151,424 @@ def testDeleteColumn(self):
...
+@pytest.fixture(scope="function")
+def city_data_from_csv():
+    csv_string = """City name, Area , Population , Annual Rainfall
+    Sydney, 2058 ,  4336374   ,      1214.8
+    Melbourne, 1566 ,  3806092   ,       646.9
+    Brisbane, 5905 ,  1857594   ,      1146.4
+    Perth, 5386 ,  1554769   ,       869.4
+    Adelaide, 1295 ,  1158259   ,       600.5
+    Hobart, 1357 ,   205556   ,       619.5
+    Darwin, 0112 ,   120900   ,      1714.7"""
```
hugovk: Please could you remove the spaces before commas, and put just one after commas?

tests/test_prettytable.py
```diff
@@ -547,10 +686,20 @@ def testAsciiBreakLine(self):
...
+        self, rows: List[List[Any]], hrule: int, expected_result: str
+    ):
+        t = PrettyTable(["Field 1", "Field 2"])
+        for row in rows:
+            t.add_row(row)
+        result = t.get_string(hrules=hrule)
+        assert result.strip() == expected_result.strip()
 
-    def testHtmlBreakLine(self):
+    def test_break_line_HTLM(self):
```
```suggestion
    def test_break_line_HTML(self):
```

tests/test_prettytable.py
```diff
@@ -125,293 +151,424 @@ def testDeleteColumn(self):
...
+#         )
+#         self.cur.execute('INSERT INTO cities VALUES ("Darwin", 112, 120900, 1714.7)')
+#         self.cur.execute('INSERT INTO cities VALUES ("Hobart", 1357, 205556, 619.5)')
+#         self.cur.execute('INSERT INTO cities VALUES ("Sydney", 2058, 4336374, 1214.8)')
+#         self.cur.execute(
+#             'INSERT INTO cities VALUES ("Melbourne", 1566, 3806092, 646.9)'
+#         )
+#         self.cur.execute('INSERT INTO cities VALUES ("Perth", 5386, 1554769, 869.4)')
+#         self.cur.execute("SELECT * FROM cities")
+#         self.x = from_db_cursor(self.cur)
```
hugovk: Remove this old commented code?

hugovk (changes requested):
Thank you for this! A few comments.

tests/test_prettytable.py
davzucky: Done

tests/test_prettytable.py
davzucky: Done

tests/test_prettytable.py
davzucky: Done

hugovk (changes requested):
Please also fix the linting:

https://github.com/jazzband/prettytable/pull/135/checks?check_run_id=3826690535

The isort can be done automatically:

`python -m pip install -U pre-commit && pre-commit run --all-files`

(And optionally `pre-commit install` to have it check your changes when you commit.)

The Flake8 things need to be done manually, I guess these are mostly things to do like `noqa: X123`.

(Another idea: it could be worth splitting the test file into smaller ones, but you don't need to do this, and probably something for a followup PR anyway.)

davzucky:
Thank you for all these information. I will add that on the Doc as well later.
Fixed the linter errors.

Yes it can be split. however the question could be how ? What could be a good per use case split.

hugovk (approved):

hugovk:
> Yes it can be split. however the question could be how ? What could be a good per use case split.

Now that's a good question... Perhaps some related test classes in the same file, and the rest stay where they are? If there's no obvious way, I wouldn't worry about it too much.

Thanks again!

PR #140, opened 2021-10-19
Add colorful tables and themes

### LINKED ISSUE #86: Add colorful tables support

BD103:
Hello! I am back again. I felt bad after leaving the original PR stale, so I came back and redid it.

(Better late than never!)

I tried to follow all the programming standards with Pre-commit and the likewise. Some of my goals for this are:

- [x] Be able to print tables colored with ANSI sequences
- [x] Custom themes
- [x] ~Re-implement previous themes~
- [x] Follow all standards enforced by pre-commit, etc.
- [x] Don't create any breaking changes (part of the reason this is a separate module)
- [x] Sufficient coverage with Pytest

Thank you for putting up with me,
~BD103

(Fixes #86)


hugovk:
I'll convert this to draft for now, and we can put it as "ready" when you're ready :)

BD103:
I still have some more tests to add, but I'm getting an error. It says that it `AttributeError: module 'pytest' has no attribute 'lazy_fixture'`. I have to test my code by ignoring the main test script, which probably isn't good.

Also, should I test if Colorama is initialized?

Thanks,
-BD103

hugovk:
> I still have some more tests to add, but I'm getting an error. It says that it `AttributeError: module 'pytest' has no attribute 'lazy_fixture'`.

You need to install pytest-lazy-fixture for this.

(It's part of `python -m pip install -e ".[tests]"`)

BD103:
Thank you! I'll get this done and report back when finished.

BD103:
I have completed my task list, so this PR is ready for review. There is probably room for improvement with the tests, but for now, it is good. :)

hugovk:
Thank you again for the PR!

Is it so that if people want to use colourful table, they need to use the `ColorTable` subclass instead of the usual `PrettyTable`?

Would it be possible to refactor it so that people could continue to use `PrettyTable` and instead set the desired theme?

BD103:
It would be possible to integrate theming into the `PrettyTable` class, but I would caution against it. `ColorTable` adds some extra processing and may cause unintended side effects. For instance:

- It initializes Colorama, which wraps `sys.stdout` and may break code related to that
- It always adds the code `\x1b[0m` to the end of each table, which will break all tests
- It assumes that you are using a theme, else unnecessary data would be outputted
  - Each character would be `VERTICAL_COLOR + VERTICAL_CHAR + RESET + DEFAULT_COLOR`, instead of `VERTICAL_CHAR`. (`\x1b[31m|\x1b[0m\x1b[32m` instead of `|`.)
  - Plus adding the logic to remove this extra data would be superfluous and complicated

While it would be very convenient in transitioning to colorful tables, it seems best to rename a class here or there.

tests/test_colortable.py
```diff
@@ -0,0 +1,97 @@
...
+        default_color="31",
+        vertical_color="32",
+        horizontal_color="33",
+        junction_color="34",
+    )
+
+
+@pytest.fixture
+def char_theme():
+    return Theme(vertical_char=":", horizontal_char="~", junction_char="*")
```
hugovk: Codecov says this line isn't covered, so not used and can be removed?

hugovk (commented):
Okay, let's try it like this! Please could you also add a small example to the README?

tests/test_colortable.py
BD103: It can, let me do that.

BD103:
There we go, the readme is updated and unused fixture removed. :)

hugovk:
Thank you!

BD103:
Great, I'm so glad we could get this merged! :D

PR #146, opened 2021-10-28
Add Jupyter repr and default interpreter repr methods

### LINKED ISSUE #127: Make Jupyter-aware with _repr_html_()

When using PrettyTable in Jupyter Notebooks, print(table) produces the ASCII table which is inferior to an HTML representation like pandas and other notebook-friendly libraries use.

A pedantic solution to using the HTML output would be:

```python
from IPython.display import HTML, display

display(HTML(table.get_html_string()))
```

But giving PrettyTable a proper `_repr_html_()` works much better: `PrettyTable._repr_html_ = PrettyTable.`

image

PrettyTable should have a `_repr_html_()` which calls `self.get_html_string()` so that it is a Jupyter-aware table.
PrettyTable should also have a basic `__repr__()` method returning the same as `__str__()` while we're at it. Not having a `repr` at all seems pretty lacking compared to most mature libraries.

amirali:
Fixes #127.

I added the features they talked about them in #127.

### DIFF
--- a/tests/test_prettytable.py
+++ b/tests/test_prettytable.py
@@ -1594,23 +1594,15 @@ class TestCustomFormatter:
         )
 
 
-@pytest.fixture
-def simple_table():
-    pt = PrettyTable(["field1", "field2"])
-    pt.add_row(["row1_field1", "row1_field2"])
-    pt.add_row(["row2_field1", "row2_field2"])
-    return pt
-
-
 class TestRepr:
-    def test_repr_html_attribute(self, simple_table: PrettyTable):
+    def test_repr_html_attribute(self, row_prettytable: PrettyTable):
         try:
-            _ = getattr(simple_table, "_repr_html_")
+            _ = getattr(row_prettytable, "_repr_html_")
         except AttributeError:
             assert False
 
-    def test_default_repr(self, simple_table: PrettyTable):
-        assert simple_table.__str__() == simple_table.__repr__()
+    def test_default_repr(self, row_prettytable: PrettyTable):
+        assert row_prettytable.__str__() == row_prettytable.__repr__()
 
-    def test_jupyter_repr(self, simple_table: PrettyTable):
-        assert simple_table._repr_html_() == simple_table.get_html_string()
+    def test_jupyter_repr(self, row_prettytable: PrettyTable):
+        assert row_prettytable._repr_html_() == row_prettytable.get_html_string()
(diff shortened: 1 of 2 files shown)


hugovk (changes requested):
Thank you! Please could you add tests to cover these lines?

amirali:
I added the test for new repr and Jupyter environment repr. Sorry I didn't add them at first

tests/test_prettytable.py
```diff
@@ -1592,3 +1592,19 @@ def my_format(col: str, value: Any) -> str:
...
+
+@pytest.fixture
+def simple_table():
+    pt = PrettyTable(["field1", "field2"])
+    pt.add_row(["row1_field1", "row1_field2"])
+    pt.add_row(["row2_field1", "row2_field2"])
+    return pt
+
+
+class TestRepr:
```
kumaraditya303, line 1597: A test for `__getattr__` with `_repr_html_` is required, Thank you

kumaraditya303 (dismissed):

tests/test_prettytable.py
amirali: Written here 921bdbe

flutefreak7:
This looks good to me! I'm the author of #127

tests/test_prettytable.py
```diff
@@ -1592,3 +1592,25 @@ def my_format(col: str, value: Any) -> str:
 +-------------+----------+-----------+------------+
 """.strip()
         )
+
+
+@pytest.fixture
+def simple_table():
```
hugovk: Can we reuse an existing fixture instead of creating a new one?

hugovk:
I'm not sure if my earlier review comments saved properly, will try again!

---

Do we need this as an attribute?

```python
        elif name == "_repr_html_":
            return self._repr_html_()
```

Looks like IPython only needs the method, not attribute:

* https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

amirali:
When I implemented the method, in the first runs in Jupyter it doesn't show the expected value in HTML and got AttriubuteError on calling the method. So I thought it should be registered as an attribute. But if you think it's not necessary I'll double-check it and remove the extra part.

hugovk:
I'm not super familiar with IPython / JupyterLab but seemed to work fine for me without the attribute:

image

Let's remove it, and if it's needed later it's easy to add. Thanks!

amirali:
Done.

hugovk (approved):

hugovk:
Thank you!

PR #147, opened 2021-10-28
Add rows attribute to prettytable

### LINKED ISSUE #128: PrettyTable should expose its data for consumption by other table libraries

If I construct a PrettyTable, it's only useful for display... it's a dead end... if I want to then use it with other libraries (like pandas, matplotlib, openpyxl, etc), I have no way to access a PrettyTable's data via the public API.

I'd like if there was a `rows` or `data` attribute exposed publicly so that the table could be an input to downstream functions.

Accessing the `table._rows` "hidden" attribute works fine, it just means I'm relying on a private implementation detail instead of a part of the API.

`df = pd.DataFrame(table._rows, columns=table.field_names)` works fine and is what I'm currently using.

kumaraditya303:
Closes https://github.com/jazzband/prettytable/issues/128

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -518,6 +518,9 @@ class PrettyTable:
     ##############################
     # ATTRIBUTE MANAGEMENT       #
     ##############################
+    @property
+    def rows(self):
+        return self._rows
 
     @property
     def xhtml(self):
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -520,7 +520,7 @@ class PrettyTable:
     ##############################
     @property
     def rows(self):
-        return self._rows
+        return self._rows[:]
 
     @property
     def xhtml(self):
--- a/tests/test_prettytable.py
+++ b/tests/test_prettytable.py
@@ -345,7 +345,7 @@ class TestBasic:
     """Some very basic tests."""
 
     def test_table_rows(self, city_data_prettytable: PrettyTable) -> None:
-        rows = city_data_prettytable.rows[:]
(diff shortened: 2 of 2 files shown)


tests/test_prettytable.py
```diff
@@ -344,6 +344,11 @@ def init_db(db_cursor):
 class TestBasic:
     """Some very basic tests."""
 
+    def test_table_rows(self, city_data_prettytable: PrettyTable) -> None:
+        rows = city_data_prettytable.rows[:]
```
hugovk: And can we skip making a copy here?
```suggestion
        rows = city_data_prettytable.rows
```

src/prettytable/prettytable.py
```diff
@@ -518,6 +518,9 @@ def _validate_attributes(self, name, val):
     ##############################
     # ATTRIBUTE MANAGEMENT       #
     ##############################
+    @property
+    def rows(self):
+        return self._rows
```
hugovk: I think my earlier review got lost, take 2:

Can we return a copy here so the original is not writable?

```suggestion
        return self._rows[:]
```

hugovk:
@kumaraditya303 It's fine this time, but please don't merge your own PRs without an approved review. Anyway, thank you for the contribution!

kumaraditya303:
> @kumaraditya303 It's fine this time, but please don't merge your own PRs without an approved review. Anyway, thank you for the contribution!

Okay, I thought I addressed your suggestion so it was okay to merge myself.

hugovk:
Yep, you did, although sometimes (usually with larger PRs) I will make partial suggestions and still want to do a final review of the whole thing, and maybe test locally, before merge.

Issue #159, opened 2022-01-05
HTML titles appear below table headers


daibhid:
### What did you do?
When a HTML table is generated and rendered in a browser, the table title appears in a row below the table headers. 

### What did you expect to happen?
The table title appears above the table headers (as it does for ASCII tables)

### What actually happened?
The web browser renders the table headers at the top, with the title underneath, as in this example: 

![pretty_table_html](https://user-images.githubusercontent.com/16548771/148298682-fbd3d081-94d3-4929-9b5d-3f98d21954a7.png)

### What versions are you using?

* OS: Ubuntu 21.10
* Python: 3.9.7
* PrettyTable: `master` (2.5.0)

Please include **code** that reproduces the issue.

### Code to reproduce issue:
This sample generates a html file `sample.html` that should be opened in a web browser to see the rendering issue

```python
from prettytable import PrettyTable
x = PrettyTable()

x.title = "Title"

x.field_names = ["City name", "Area", "Population", "Annual Rainfall"]
x.add_row(["Adelaide", 1295, 1158259, 600.5])
x.add_row(["Brisbane", 5905, 1857594, 1146.4])
x.add_row(["Darwin", 112, 120900, 1714.7])
x.add_row(["Hobart", 1357, 205556, 619.5])
x.add_row(["Sydney", 2058, 4336374, 1214.8])
x.add_row(["Melbourne", 1566, 3806092, 646.9])
x.add_row(["Perth", 5386, 1554769, 869.4])

print(x)
print(x.get_html_string())

with open("sample.html", "w") as f:
    f.write(x.get_html_string())
```

Note that the HTML tags are generated in the correct order, but web browsers render the table in a different order

hugovk:
Generated HTML:

```html
<table>
    <tr>
        <td colspan=4>Title</td>
    </tr>
    <thead>
        <tr>
            <th>City name</th>
            <th>Area</th>
            <th>Population</th>
            <th>Annual Rainfall</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>Adelaide</td>
            <td>1295</td>
            <td>1158259</td>
            <td>600.5</td>
        </tr>
        <tr>
            <td>Brisbane</td>
            <td>5905</td>
            <td>1857594</td>
            <td>1146.4</td>
        </tr>
        <tr>
            <td>Darwin</td>
            <td>112</td>
            <td>120900</td>
            <td>1714.7</td>
        </tr>
        <tr>
[...]

PR #160, opened 2022-01-05
Use `<caption>` tags to print html table titles

daibhid:
This is a potential fix for issue #159

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2005,12 +2005,7 @@ class PrettyTable:
         # Title
         title = options["title"] or self._title
         if title:
-            cols = (
-                len(options["fields"]) if options["fields"] else len(self.field_names)
-            )
-            lines.append("    <tr>")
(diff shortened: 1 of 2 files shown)


hugovk:
Thank you, `<caption>` looks much more suitable. It's also possible for the user to style it with CSS to choose which side to put it. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption

Please could you add tests?

hugovk:
Looks good, would it also be possible to add one for the change in `_get_formatted_html_string` as well? I realise neither of these were tested/covered before, but now's the perfect opportunity to improve that.

daibhid:
> Looks good, would it also be possible to add one for the change in `_get_formatted_html_string` as well? I realise neither of these were tested/covered before, but now's the perfect opportunity to improve that.

Yes, I'd be happy to, will update this soon

hugovk:
Thank you for the report and pull request!

hugovk:
Released in PrettyTable 3.0.0, thanks again!

PR #162, opened 2022-01-25
Added requirements.txt to clarify installation requirements

vivian-hafener:
While packaging the prettytable module in spack, the installation errored out due to the lack of the depencency wcwidth. I've PR'd this requirements.txt so that people packaging or installing this module in the future are aware of this dependency.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+wcwidth
+
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,1 @@
 wcwidth
-


hugovk:
Thanks for the PR!

I'd prefer not to add a `requirements.txt` because the dependencies are already defined in `setup.cfg` and I'd rather not duplicate them:

https://github.com/jazzband/prettytable/blob/03f6ce926cdb22150e676daef39210f91f4201a3/setup.cfg#L28-L30

As we can see, the propsed `requirements.txt` is missing the conditional `importlib` dependency.

I'm not familiar with Spack, but I would suggest it also reads setuptools' `install_requires` metadata, as many other libraries also only use `install_requires` to specify minimal dependencies.

https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#install-requires

Does this help? https://spack.readthedocs.io/en/latest/build_systems/pythonpackage.html#setuptools

vivian-hafener:
Ok, that makes sense! My apologies for the extraneous PR

PR #164, opened 2022-02-05
Convert None to empty cell or custom value

### LINKED ISSUE #163: Display null / None values as an empty cell?
I would like to make a table with `prettytable` that displays null / None values as empty cells, like this (table made using `tabulate` library)

Screenshot from 2022-01-27 18-14-15

As opposed to displaying the `None` values literally (this one uses `prettytable`)

Screenshot from 2022-01-27 18-14-11

Is this currently possible?  And if not, it would be a great feature

av-guy:
I added the ability to format None values on a column-by-column basis.

resolves #163 

Please forgive me if I have failed to do something properly. This is my first contribution to an open source project. Criticisms are welcome.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -72,7 +72,6 @@ def _get_size(text):
 
 class PrettyTable:
     def __init__(self, field_names=None, **kwargs):
-
         """Return a new PrettyTable instance
 
         Arguments:
@@ -188,6 +187,7 @@ class PrettyTable:
             "valign",
             "max_width",
             "min_width",
+            "none_format"
         ]
         for option in self._options:
             if option in kwargs:
@@ -199,6 +199,7 @@ class PrettyTable:
         self._start = kwargs["start"] or 0
         self._end = kwargs["end"] or None
         self._fields = kwargs["fields"] or None
+        self._none_format = {}
 
         if kwargs["header"] in (True, False):
             self._header = kwargs["header"]
@@ -227,6 +228,7 @@ class PrettyTable:
         self.int_format = kwargs["int_format"] or {}
         self.float_format = kwargs["float_format"] or {}
         self.custom_format = kwargs["custom_format"] or {}
+        self.none_format = kwargs["none_format"] or {}
 
         self._min_table_width = kwargs["min_table_width"] or None
         self._max_table_width = kwargs["max_table_width"] or None
@@ -342,6 +344,8 @@ class PrettyTable:
     def _validate_option(self, option, val):
         if option == "field_names":
             self._validate_field_names(val)
+        elif option == "none_format":
+            self._validate_none_format(val)
         elif option in (
             "start",
             "end",
@@ -424,6 +428,15 @@ class PrettyTable:
             assert len(val) == len(set(val))
         except AssertionError:
             raise Exception("Field names must be unique")
+
+    def _validate_none_format(self, val):
+        try:
+            if val is not None:
+                assert isinstance(val, str)
+        except AssertionError:
+            raise Exception(
+                "Replacement for None value must be a string if being supplied."
+            )
 
     def _validate_header_style(self, val):
         try:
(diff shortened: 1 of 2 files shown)


tests/test_prettytable.py
```diff
@@ -94,6 +94,60 @@ def mix_prettytable():
     return mix
 
 
+class TestNoneOption:
+    def test_none_char_valid_option(self):
+        t = PrettyTable(["Field 1", "Field 2", "Field 3"], none_format="")
```
hugovk: This `t` variable isn't used, so we can skip it (also silences the lint CI): 

```suggestion
        PrettyTable(["Field 1", "Field 2", "Field 3"], none_format="")
```

tests/test_prettytable.py
```diff
@@ -94,6 +94,60 @@ def mix_prettytable():
...
+
+    def test_none_char_invalid_option(self):
+        with pytest.raises(Exception) as exc:
+            t = PrettyTable(["Field 1", "Field 2", "Field 3"], none_format=2)
```
hugovk: Same here:
```suggestion
            PrettyTable(["Field 1", "Field 2", "Field 3"], none_format=2)
```

tests/test_prettytable.py
```diff
@@ -94,6 +94,60 @@ def mix_prettytable():
...
+        assert (
+            """
++---------+---------+---------+
+| Field 1 | Field 2 | Field 3 |
++---------+---------+---------+
+| value 1 |   None  | value 2 |
++---------+---------+---------+
+""".strip()
+            == t.get_string().strip()
+        )
```
hugovk: Little nit: pytest asserts are usually written in the order `assert actual == expected`.

Doesn't affect the True/False, but when things fail the error message should be a bit clearer.

Please could you put `t.get_string().strip()` before the `==` and the expected table afterwards? And the same with the next two tests.

Thanks!

src/prettytable/prettytable.py
```diff
@@ -1801,6 +1830,8 @@ def _stringify_row(self, row, options, hrule):
             lines = value.split("\n")
             new_lines = []
             for line in lines:
+                if line == "None" and self.none_format[field] is not None:
```
hugovk, line 1833: So if the cell contains either `None` or the string `"None"`, it'll be replaced (e.g. by "N/A").

Is this intentional and/or acceptable?

(I don't have a strong preference either way, just wanted to check.)

hugovk (commented):
Thank you for the PR!

For a first contribution to open source, this is a very good job!

It's really good you included tests which cover the different aspects of the new feature.

Just a few small suggestions and comments :)

src/prettytable/prettytable.py
av-guy: Hi Hugo,

Thanks for the kind comments and for making this process less intimidating than I thought it would be.

Yes, that was the intended behavior. The idea was to allow the user to replace any "None" values with a value of their choosing. So in my tests I used "N/A" (not available) to represent a None value.

After some thought, I don't know if the way I did it is best. I am thinking there could be a possibility where, after `lines = value.split("\n")`, there could be a "None" value in the split text for some other reason.

Do you think a better evaluation would be to see if `lines` has only a single value of "None" and then replace "None" with the `none_format` value?

hugovk: I think the way you've done it is fine, let's also include it in a test case (I'll add a suggestion below).

Thanks!

tests/test_prettytable.py
```diff
@@ -94,6 +94,60 @@ def mix_prettytable():
...
+
+    def test_replace_none_all(self):
+        t = PrettyTable(["Field 1", "Field 2", "Field 3"], none_format="N/A")
+        t.add_row(["value 1", None, None])
```
hugovk: Let's include both a `None` and `"None"` string here to demonstrate them both being replaced by `"N/A"`:

```suggestion
        t.add_row(["value 1", None, "None"])
```

tests/test_prettytable.py
av-guy: https://github.com/jazzband/prettytable/pull/164/commits/fa72018366f15b2c998c611541636c568138ca6f

I made the changes you requested on the above commit. Please let me know if there is anything else I need to do.

hugovk:
Thank you for your contribution!

PR #169, opened 2022-02-23
Use concrete built-in exceptions instead of Exception base class

### LINKED ISSUE #168: use custom exception type
no

use a custom exception/error type so one does not have to catch all errors of type `Exception`

A clear and concise description of any alternative solutions or features you've
considered.

hugovk:
Fixes https://github.com/jazzband/prettytable/issues/168.

Instead of using the `Exception` base class, use concrete built-in `IndexError`, `TypeError` and `ValueError` instead.

https://docs.python.org/3/library/exceptions.html

Allows more targeted exception catching.


fawaf:
:shipit:

Issue #172, opened 2022-03-05
Width seems to be wrong with custom none_format


OlafvdSpek:
```python
from prettytable import PrettyTable
t = PrettyTable()
t.add_row([None])
t.none_format='0123456789'
print(t)
t.header = False
print()
print(t)
```
```
+---------+
| Field 1 |
+---------+
| 0123456 |
|   789   |
+---------+

+------+
| 0123 |
| 4567 |
|  89  |
+------+
```

hugovk:
### What did you expect to happen? 

### What versions are you using?

* OS: 
* Python: 
* PrettyTable:

OlafvdSpek:
Expected:
```
+------------+
|  Field 1   |
+------------+
| 0123456789 |
+------------+

+------------+
| 0123456789 |
+------------+
```
OS: Windows 10 and Debian unstable
Python 3.10 and 3.9
PrettyTable: git and 3.1.1

hugovk:
Thanks!

And for comparison, adding a single row with the string `'0123456789'` also produces single-line output.

```python
from prettytable import PrettyTable
t = PrettyTable()
# t.add_row([None])
# t.none_format='0123456789'
t.add_row(['0123456789'])
print(t)
t.header = False
print()
print(t)
```
```
+------------+
|  Field 1   |
+------------+
| 0123456789 |
+------------+

+------------+
| 0123456789 |
+------------+
```

`none_format` was recently added in https://github.com/jazzband/prettytable/pull/164.

@av-guy Would you like to take a look at this? Thanks!

av-guy:
Yes I will review immediately and see what's going on.

av-guy:
@hugovk I was able to figure out what is causing this issue, and I will be making a PR for the fix later this evening.

hugovk:
Thanks both, fixed in #174.

PR #174, opened 2022-03-06
Fix width for custom `none_format`

av-guy:
Resolves issue #172 by checking for existence of a `None` value when widths are being computed inside of `_compute_widths`.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1510,6 +1510,9 @@ class PrettyTable:
         for row in rows:
             for index, value in enumerate(row):
                 fieldname = self.field_names[index]
+                if self.none_format.get(fieldname) is not None:
+                    if value == 'None' or value is None:
+                        value = self.none_format.get(fieldname)
                 if fieldname in self.max_width:
                     widths[index] = max(
                         widths[index],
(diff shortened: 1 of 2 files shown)


src/prettytable/prettytable.py
```diff
@@ -1836,6 +1839,8 @@ def _stringify_row(self, row, options, hrule):
             for line in lines:
                 if line == "None" and self.none_format.get(field) is not None:
                     line = self.none_format[field]
+                    if len(line) > width:
+                        width = len(line)
```
hugovk: This line isn't covered by tests, is possible to add one?

av-guy: I think the easiest way to test that would be to move it to its own function. In fact, maybe I should just move that entire statement to its own function and test the function individually. Do you think that would be okay?

hugovk: In this case, as it's just a couple of lines, I think it'd be better to test via the public API so we know that this code is actually needed.

What sort of input would trigger this line?

av-guy: I think the check is redundant, and I am pretty sure it can be removed. I did so on my end and all of the tests still pass. The width is recalculated inside of `_compute_widths`.

hugovk: Thanks for checking, let's remove it then!

av-guy: Should now be fixed with 05f0d707.

hugovk:
Thank you!

Issue #148, opened 2021-10-30
Addtional horizontal lines to create sections within a table


borderite:
### Is your feature request related to a problem? Please describe.

I often want to make groups of rows in a table. To indicate the grouping, I want to put a horizontal line between each pair of the adjacent row groups.

### Describe the solution you'd like

It would be great if I could place a horizontal line after any row I pick (or before any row I pick).

### Describe alternatives you've considered

The python package, rich, let me do what I would like to do (by using "end_section") when I want text outputs. But rich does not offer a method like get_latex_string.

### Additional context

Here is an example of text outputs I desire to get.  The horizontal line right above "Summary Statistics" is the one I cannot get using PrettyTable.

```
        Dependent variable: birthwdight       
=============================================
                     (1)      (2)      (3)   
---------------------------------------------
Regressor                                    
  smoker            -253.23  -217.58  -175.38
                   (26.801)  (26.09) (26.804)
  alcohol                    -30.491  -21.083
                            (72.548) (72.931)
  nprevist                     34.07   29.603
                            (3.6059) (3.5798)
  unmarried                           -187.13
                                     (27.654)
  const              3432.1   3051.2   3134.4
                   (11.887) (43.685) (44.112)
---------------------------------------------
Summary Statistics                           
  SER                583.73   570.47    565.7
  R-squared        0.028604  0.07285 0.088605
  Adj. R-squared    0.02828 0.071922 0.087388
---------------------------------------------
```

kinredon:
How to use this function?

hugovk:
@kinredon It hasn't been implemented. I would welcome a PR, something similar to Rich's `end_section` sounds pretty good.

https://rich.readthedocs.io/en/stable/reference/table.html#rich.table.Row

hugovk:
This is now possible with the `add_divider()` method or `add_row(row, divider=True)`:

https://github.com/prettytable/prettytable#adding-sections-to-a-table

PR #185, opened 2022-05-20
Add horizontal lines to create sections within a table

myheroyuki:
I wanted to try to add the ability to create sections in a table as requested in issue #148. In order to add an `end_section` attribute to a row, I changed table rows from a simple list to a class. It doesn't change anything from the user's perspective, but it's a bit of a big change to the code, so if there's anything you'd like me do differently I'm open to any suggestions.


hugovk:
Thanks for the PR!

> It doesn't change anything from the user's perspective, ...

I can see one change:

```diff
-                assert len(val) == len(self._rows[0])
+                assert len(val) == len(self._rows[0].data)
```

Before `x.rows` would give a list of lists of the table:
```
[['Adelaide', 1295, 1158259, 600.5], ['Brisbane', 5905, 1857594, 1146.4], ['Darwin', 112, 120900, 1714.7], ['Melbourne', 1566, 3806092, 646.9], ['Perth', 5386, 1554769, 869.4], ['Sydney', 2058, 4336374, 1214.8]]
```
But now it gives:
```
[<prettytable.prettytable.TableRow object at 0x1025d2d10>, <prettytable.prettytable.TableRow object at 0x1025d0070>, <prettytable.prettytable.TableRow object at 0x10258ff40>, <prettytable.prettytable.TableRow object at 0x1028c2e00>, <prettytable.prettytable.TableRow object at 0x1028c2a10>, <prettytable.prettytable.TableRow object at 0x1028c29b0>, <prettytable.prettytable.TableRow object at 0x1028c2950>]
```

This is potentially a breaking change.

myheroyuki:
Good catch, I should have seen that. I suppose I could edit the `@property` method for `rows` to be something like

`return [row.data[:] for row in self._rows]`.

That seems to work fine from the testing I did, but I don't know that I like that as a solution since it would be returning something different than the actual `rows` value. 

Other than that I don't see any good way to keep the return value of `x.rows` consistent while adding a new attribute to the row. The new `end_section` attribute could be set by row index instead of being part of the row itself, but that might get a bit messy when things like sorting change the index of the row. Maybe a new internal structure could be made to map each row to its attributes? I'll give this a bit more thought and see if I can find a clean solution.

hadrielk:
I'm not positive, but wouldn't the way this change was written essentially make sorting useless, if `end_section` is being used?

Because internally prettytables isn't sorting only within a bounded range of such marked "sections", so it will sort across the whole table and mess it up. No?

Also, does the `sort_key` still work the same way as before? It looks like the invoked function is now being passed only the `sortby`'s field, whereas before it was given a `list`, with the `sortby` being the first entry in that list and the rest being the row's fields. So that's a breaking change too, no?

---

As an aside, at my day job I changed our fork of prettytable to do something like this, but only specifically for "trailing row(s)", as is needed for totals/summary at the end of the whole table. That's even what #148 shows as an example, because that's a very common use case compared to multiple sections. And doing that may be easier. (it was fairly straightforward for me, just storing it as a separate list _not_ in `self._rows` and adding them with a separate `add_trailer()` method instead of `add_row()`... but we only use prettytable for sorting and ascii output so ymmv)

myheroyuki:
@hadrielk yeah, sorting with `end_section` would probably be break the utility of the sections in most cases, but I can think of some cases where it might still be something a user would want. I'm not sure what the best default behavior for this situation would be. Should the sections be removed if the table is sorted, or should the section lines stay unless the user removes them in case they still want them? I suppose either option is okay as long as it's properly documented. 

I think re-implementing the feature similarly to how you did it is probably the best option in the long run for not breaking any backwards compatibility.

src/prettytable/prettytable.py
```diff
@@ -1664,6 +1670,24 @@ def _get_rows(self, options):
...
+        options - dictionary of option settings."""
+
+        if options["oldsortslice"]:
+            dividers = copy.deepcopy(self._dividers[options["start"] : options["end"]])
+        else:
+            dividers = copy.deepcopy(self._dividers)
+
+        if options["sortby"]:
+            for i in range(len(dividers)):
+                dividers[i] = False
```
```suggestion
            dividers = [False for divider in dividers]
```

src/prettytable/prettytable.py
```diff
@@ -1431,7 +1432,7 @@ def add_rows(self, rows):
         for row in rows:
             self.add_row(row)
 
-    def add_row(self, row):
+    def add_row(self, row, divider=False):
```
hugovk: Because it can be a bit unclear what Boolean means in `add_row(row, True)` or `add_row(row, False)`, shall we use a keyword argument (aka named parameter) for this new one?

```suggestion
    def add_row(self, row, *, divider=False):
```

Then we need to explicitly do `add_row(row, divider=True)` or `add_row(row, divider=False)` instead, which is much clearer.

src/prettytable/prettytable.py
```diff
@@ -1463,6 +1465,7 @@ def del_row(self, row_index):
                 f"table only has {len(self._rows)} rows"
             )
         del self._rows[row_index]
+        del self.dividers[row_index]
```
hugovk: This line isn't covered by tests. Please can you add a test case? Also for lines 1539 and 1547 below.

hugovk (changes requested):
Thanks for the update!

myheroyuki:
Thanks for the quick review. Let me know if there's anything else you want me to change.

hadrielk:
I have another question: does this work for HTML and Latex outputs?

Because if not, and your main use-case is ascii output, then I have an alternative proposal: don't do this inside prettytable. Multiple sections (and trailers too) are logically equivalent to just multiple tables, one after another.

The only reason we need prettytable to do it for us is we don't know the table and individual column widths to use for the different sections. But we could expose an API from prettytable to let it tell us that info, in a convenient fashion.

It already has the function internally to determine that info for its own member table, that it uses when it prints it. So we could expose something similar - a `@staticmethod` that accepts an iterable of `prettytable`s, and returns a namedtuple of `(table_width, [col_width])`. That tuple could then be passed into `get_string()` as keyword argument, such that prettytable uses the tuple's pre-calculated widths for printing its table.

In other words, from the user's perspective, they can do this:
```python
FIELD_NAMES=["City name", "Area", "Population", "Annual Rainfall"]

x = PrettyTable()
x.field_names = FIELD_NAMES
x.header = True
x.add_row(["Adelaide", 1295, 1158259, 600.5])
x.add_row(["Brisbane", 5905, 1857594, 1146.4])

y = PrettyTable()
y.field_names = FIELD_NAMES
y.header = False
y.add_row(["Darwin", 112, 120900, 1714.7])
y.add_row(["Hobart", 1357, 205556, 619.5])

totals = PrettyTable()
totals.field_names = FIELD_NAMES
totals.header = False
totals.add_row(["Totals", 9504, 12345678, 4442.7])

widths = PrettyTable.get_widths([x, y, totals])

print('\n'.join([section.get_string(widths=widths) for section in (x, y, totals)]))
```

And from `prettytable`'s perspective, it's a small change, I think?

hadrielk:
The advantages to doing it that way, btw, is:

1. It gives the user more control
   - They can sort each section independently.
   - They can decide what section to print in which order.
   - They can decide the formatting of each section, independently.
2. It keeps `prettytable`'s internal code relatively clean and with few changes (I think).

myheroyuki:
@hadrielk This doesn't currently work for HTML or Latex outputs, but the plan was to add that functionality in the future. I wanted to submit a pull request for the base functionality to get some feedback on how the feature was implemented before putting in the time to make it work for every output format. After seeing the comments on the PR it seems that was the right choice as I completely changed the original implementation. 

What you suggested makes sense though, and I can see use cases where doing what you described would be preferable to adding section dividers within a single table.

romanbsd:
I hope to see this feature :)

hugovk:
Thanks for the ping! I think this is almost ready for merge.

@myheroyuki Please could you resolve the conflicts and I'll give it another review? This will likely also need type hints, we can see what mypy says.

hugovk:
I've just updated `main` to fix the lint failures for isort/poetry. Please could you update again? Thanks!

src/prettytable/prettytable.py
```diff
@@ -134,8 +134,10 @@ def __init__(self, field_names=None, **kwargs) -> None:
         self.encoding = kwargs.get("encoding", "UTF-8")
 
         # Data
+
```
```suggestion
```

myheroyuki:
I think everything is looking good now. 
@hugovk, Let me know if you have any other suggestions. If not I'll merge this in

hugovk (approved):
One little nit, but otherwise looks good, thanks!

hugovk:
Thank you!

con-f-use:
Just out of curiosity, if I wanted to add a divider after the fact, to a pre-existing table. Say between row 6 and 7, how would I go about it in that code? I would guess `table._dividers[6] = True` right? Probably not the best API for that usecase.

Also, the main page of this github repo should probably not display the latest README, because it makes it seem that `divider` is already part of the current release (3.6.0), but it is not.

hugovk:
> Also, the main page of this github repo should probably not display the latest README, because it makes it seem that `divider` is already part of the current release (3.6.0), but it is not.

What should it show?

If we change the default branch from `main` to something else, like some moving `released` branch, that means PRs will go to that (wrong) branch instead of `main` and also complicate the release process.

con-f-use:
You are right, I somehow thought I had seen other projects use a "display branch" separate from the default branch, with the default branch being the one that gets checked out when cloning, but that is not possible.
Maybe a very prominent link to a readthedocs documentation. "You are viewing the development branch, for latest stable documentation go here." in the readme type of thing?

hugovk:
Yeah, often readthedocs will have a latest and also a stable build, but not so with GitHub repos. We also don't have any readthedocs documentation to link to :)

Anyway, I've just released this in 3.7.0, so the README is now up-to-date.

con-f-use:
Thanks for your work, I really appreciate it and the people who maintain FOSS projects.

Btw. another idea would be do the "You are viewing the development branch, for latest stable documentation go here." think, but the link points to the release commit of the Readme instead of readthedocs. Don't know if it's worth it because it adds a bit of process, but that should be easily automatable.

Also sorry for the offtopic, a more public (and documented) API for setting dividers on a table with pre-existing rows would still be nice.

PR #186, opened 2022-06-01
`None` `annotation` for `__init__` constructors/functions

ShobanChiddarth:
Everywhere there was a class declaration and a __init__ function
defined for that class, I added `None` as an `annotation` to it.
```python
class ...:
    def __init__(..) -> None:
```

[Reason](https://peps.python.org/pep-0484/#the-meaning-of-annotations)

### DIFF
--- a/src/prettytable/colortable.py
+++ b/src/prettytable/colortable.py
@@ -23,7 +23,7 @@ class Theme:
         horizontal_color: str = "",
         junction_char: str = "+",
         junction_color: str = "",
-    ):
+    ) -> None:
         self.default_color = Theme.format_code(default_color)
         self.vertical_char = vertical_char
         self.vertical_color = Theme.format_code(vertical_color)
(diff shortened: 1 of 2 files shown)


ShobanChiddarth:
image
3 Checks have failed. Is there anything wrong? Is there anything I still have to change?

hugovk:
That's a bug in the new Python 3.11 beta 2, see:

* https://github.com/pytest-dev/pytest/issues/10008
* https://github.com/python/cpython/issues/92597
* https://github.com/python/cpython/pull/93359
* https://github.com/python/cpython/pull/93397
* https://mail.python.org/archives/list/python-dev@python.org/thread/VZ222BFFAWW73DYFCR7NZR2C5M3TXD65/

There's a workaround, but they may be releasing beta 3 soon to fix it.

Anyway, it's not a problem for this PR, let's just merge it. Thanks for the contribution!

ShobanChiddarth:
@hugovk
> Anyway, it's not a problem for this PR, let's just merge it. Thanks for the contribution!

Welcome. Regards.

PR #193, opened 2022-08-17
Add option to specify line break character used in pagination

myheroyuki:
In some cases when getting multiple pages of table data it's preferable to have the pages divided by breaks other than form feed. This allows users to separate pages by whatever they'd like.

### DIFF
--- a/tests/test_prettytable.py
+++ b/tests/test_prettytable.py
@@ -1735,6 +1735,9 @@ def test_paginate():
     assert paginated.startswith(expected_page_1)
     assert "\f" in paginated
     assert paginated.endswith(expected_page_2)
+    paginated = t.paginate(page_length=4, line_break="\n")
+    assert "\f" not in paginated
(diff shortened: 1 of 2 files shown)


tests/test_prettytable.py
```diff
@@ -1735,6 +1735,9 @@ def test_paginate():
     assert paginated.startswith(expected_page_1)
     assert "\f" in paginated
     assert paginated.endswith(expected_page_2)
+    paginated = t.paginate(page_length=4, line_break="\n")
+    assert "\f" not in paginated
+    assert "\n" in paginated
```
hugovk, line 1744: Please either: move these new test to its own test case (can use a fixture or helper function to set up the "Arrange" stuff, or add new lines and new "Act" and "Assert" comments.

hugovk (commented):
Looks good, one suggestion for the test. Thanks!

hugovk:
Thanks!

Issue #203, opened 2022-10-03
Add type hints from typeshed


hugovk:
Let's add type hints to PrettyTable.

The good news is much of the work is already done because typeshed has .pyi signature stubs:

* https://github.com/python/typeshed/tree/master/stubs/prettytable/prettytable

And here are some resources for automating the conversion of stubs into inline annotations:

* https://typing.readthedocs.io/en/latest/#type-hint-and-stub-integration

Tasks:

* [x] Add this to `.pre-commit-config.yaml` so pre-commit adds this import to Python files and we can use newer typing Syntax on older Python 3.7:

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 3dbc9c0..316e7f5 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,6 +15,7 @@ repos:
     rev: 5.10.1
     hooks:
       - id: isort
+        args: [--add-import=from __future__ import annotations]
 
   - repo: https://github.com/PyCQA/flake8
     rev: 5.0.4
```

* [x] Apply the type hints from typeshed to this repo
* [ ] Add mypy via pre-commit and make mypy pass (can be follow-up PR if there's much to do)
* [x] Add py.typed and "Typing :: Typed" Trove classifier

phershbe:
I can work on this if you want. I'm relatively new but I understand type hinting. I know the basics of how to run Mypy against type hints, but I am working on understanding Typeshed for this issue. Obviously the links explain how to kind of automate it with the .pyi signature stubs so that I don't have to type out the hint for each specific case ... I'm reading the documentation from the links now. I can work hard on it and hopefully be done really quickly.

hugovk:
Sounds good, thank you! I've not used those tools either, but feel free to ask if you've any questions and we'll work it out :)

phershbe:
@hugovk I'm working on this still. I didn't even know what stubs or .pyi files were until a few days ago; some PEPs explain them and your links above do too. I'm here to help the project not hinder progress though, so if you or anybody else want to submit the PR, go for it. I'm still here and working on it too!

hugovk:
Thank you, if you want to keep working on it that's fine, no rush!

phershbe:
[...]

PR #205, opened 2022-10-14
Add type hints from typeshed

phershbe:
Issue: https://github.com/jazzband/prettytable/issues/203

I didn't do the third part yet...
**Add mypy via pre-commit and make mypy pass (can be follow-up PR if there's much to do)**
...as I'm newer to contributing and didn't want to get overwhelmed. I can open a new issue so that it doesn't get forgotten and work on that next.

Used merge_pyi to integrate the type annotations and added appropriate args to .pre-commit-config.yaml in order for pre-commit to add the import to Python files so that newer typing syntax can be used on older Python 3.7 (as described in the issue).

Note: Did not use Black to format as it was deleting spaces in the docstrings which presumably are supposed to be there; can use Black and then add the spaces again after if that's better.

### DIFF
--- a/src/prettytable/colortable.py
+++ b/src/prettytable/colortable.py
@@ -1,4 +1,8 @@
 from .prettytable import PrettyTable
+from _typeshed import Incomplete
+from collections.abc import Callable
+from typing import ClassVar
+from .prettytable import PrettyTable
 
 try:
     from colorama import init
@@ -68,7 +72,7 @@ class ColorTable(PrettyTable):
         self._theme = value
         self.update_theme()
 
-    def update_theme(self):
+    def update_theme(self) -> None:
         theme = self._theme
 
         self._vertical_char = (
@@ -92,5 +96,5 @@ class ColorTable(PrettyTable):
             + theme.default_color
         )
 
-    def get_string(self, **kwargs):
+    def get_string(self, **kwargs) -> str:
         return super().get_string(**kwargs) + RESET_CODE
--- a/src/prettytable/colortable.py
+++ b/src/prettytable/colortable.py
@@ -1,10 +1,5 @@
 from __future__ import annotations
 
-from collections.abc import Callable
-from typing import ClassVar
-
-from _typeshed import Incomplete
-
 from .prettytable import PrettyTable
 
 try:
--- a/src/prettytable/colortable.py
+++ b/src/prettytable/colortable.py
@@ -1,7 +1,10 @@
-from .prettytable import PrettyTable
-from _typeshed import Incomplete
+from __future__ import annotations
+
 from collections.abc import Callable
 from typing import ClassVar
+
+from _typeshed import Incomplete
+
 from .prettytable import PrettyTable
 
 try:
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -46,7 +46,6 @@ from html.parser import HTMLParser
 from typing import Any
 
 import wcwidth
-from _typeshed import Incomplete
 
 # hrule styles
 FRAME = 0
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -31,6 +31,8 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+from __future__ import annotations
+
 import copy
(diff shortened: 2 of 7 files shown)


src/prettytable/prettytable.py
```diff
@@ -41,8 +43,10 @@
 import textwrap
 from html import escape
 from html.parser import HTMLParser
+from typing import Any
 
 import wcwidth
+from _typeshed import Incomplete
```
hugovk: We don't have access to this `_typeshed` module so need something else.

It's an alias for `typing.Any`:

https://github.com/python/typeshed/blob/f9ed090cfc054fa7393ad1affb30e76f238871f8/stdlib/_typeshed/__init__.pyi#L32-L37

Generally it's best to avoid `Any` and use something more specific, but for a first iteration let's use `Any` (already imported above).

```suggestion
```

src/prettytable/colortable.py
```diff
@@ -1,3 +1,10 @@
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import ClassVar
+
+from _typeshed import Incomplete
+
```
hugovk: These aren't used in this file, let's remove them:

```suggestion
```

src/prettytable/prettytable.py
```diff
@@ -558,11 +562,11 @@ def _validate_attributes(self, name, val):
     # ATTRIBUTE MANAGEMENT       #
     ##############################
     @property
-    def rows(self):
+    def rows(self) -> list[Incomplete]:
```
```suggestion
    def rows(self) -> list[Any]:
```

hugovk (commented):
Thank you! 

> I didn't do the third part yet... **Add mypy via pre-commit and make mypy pass (can be follow-up PR if there's much to do)** ...as I'm newer to contributing and didn't want to get overwhelmed. I can open a new issue so that it doesn't get forgotten and work on that next.

Yep, let's keep the scope narrower for this first PR. 👍 

Here's some review comments.

hugovk:
Now we have `from __future__ import annotations` on the files, we're using the newer syntax even for older Python versions like 3.7.

That means in `tests/test_prettytable.py`, we no longer need the existing `from typing import List` and can refer to plain `list` instead of `List` further down.

So please also make these changes:


```diff
-from typing import Any, List
+from typing import Any
```

```diff
     def test_break_line_ASCII(
-        self, rows: List[List[Any]], hrule: int, expected_result: str
+        self, rows: list[list[Any]], hrule: int, expected_result: str
     ):
```

phershbe:
@hugovk Awesome, thank you, this feedback is invaluable to me. I was sick on the weekend so I'm working on this now, I'll have it done today.

phershbe:
> Now we have `from __future__ import annotations` on the files, we're using the newer syntax even for older Python versions like 3.7.
> 
> That means in `tests/test_prettytable.py`, we no longer need the existing `from typing import List` and can refer to plain `list` instead of `List` further down.
> 
> So please also make these changes:
> 
> ```diff
> -from typing import Any, List
> +from typing import Any
> ```
> 
> ```diff
>      def test_break_line_ASCII(
> -        self, rows: List[List[Any]], hrule: int, expected_result: str
> +        self, rows: list[list[Any]], hrule: int, expected_result: str
>      ):
> ```

Done. If I'm not mistaken, the change in `test_break_line_ASCII` must have already been done automatically when I used merge_pyi.

phershbe:
@hugovk I made the changes and changed the request from `draft PR` to `PR`. I'm a little bit nervous to break something since I don't have much experience, does the pre-commit check that more or less?

I also created an issue at https://github.com/jazzband/prettytable/issues/206 for the unfinished part of the issue that this pull request is based upon so that there would be no way for it to get ignored, even though I'm starting to work on it now.

hugovk:
Thank you!

Yes, pre-commit runs tools like linters and formatters to check the code is formatted well, and for some types of errors (this is defined in `.pre-commit-config.yaml`).

Plus a test suite is run, which runs the unit tests with six versions of Python on three operating systems (they're in the `tests` directory).

And I'll review the changes to make sure they look good. 👍

hugovk (approved):
Thank you for your contribution!

phershbe:
@hugovk wow I can't thank you enough for how professional you've been answering all of my questions. Getting started making contributions to actual projects is intimidating, and having help like this is awesome. I will pay it forward and help other people when I get much better.

hugovk:
You're very welcome, I'm happy to help out :)

Issue #202, opened 2022-09-30
Alignment not validated when set for specific field


myheroyuki:
### What did you do?
Attempted to set the alignment for a header field to the  invalid alignment value "rice".

### What did you expect to happen?

ValueError: Alignment rice is invalid, use l, c or r. This is what happens when the alignment is set globally and not just for a single field.

### What actually happened?

No error message, and the table printed out with the default alignment. If I try to print a formatted html table it crashes with a key error.
It seems like the setter where the value is validated isn't called when assigning a key a value in the align dict.

Please include **code** that reproduces the issue.

The [best reproductions](https://stackoverflow.com/help/minimal-reproducible-example)
are
[self-contained scripts](https://ericlippert.com/2014/03/05/how-to-debug-small-programs/)
with minimal dependencies.

```python
t = helper_table(rows=3)
t.format = True
t.align["Field 1"] = "rice"
print(t.get_string())
```

This is another small alignment issue I found looking into other alignment related issues like #199,  #102 and #111

stuertz:
Currently there is no support vor attribute setting that way. You currently could

```python
t.align={"Field 1": "rice"}
```

And would get the expected error.

I'll look deeper into that and will add support for attr setting of all column specific options.

Issue #199, opened 2022-09-21
Align Issue with add_autoindex


myheroyuki:
When trying to fix issues with `header_align` I noticed an existing issue with how the alignment is set in `add_autoindex`. The alignment for the index column is set based on the existing values of `self.align` and `self.valign`. This would be fine if those values were strings, but they can be dicts as well.  In this case the `align[<index_fieldname>]` value gets set to a dict, creating an issue with embedded dictionaries.

### What did you do?
I created a table with align values and then added an auto_index column.

### What did you expect to happen?

I expected the index column to be properly aligned.

### What actually happened?

The index column was not properly aligned when calling `get_string`. When calling `get_html_string` the program crashed with an `TypeError: unhashable type: 'dict'` error.

Please include **code** that reproduces the issue.

```python
import prettytable
from prettytable import PrettyTable


def helper_table(rows=3):
    t = PrettyTable(["Field 1", "Field 2", "Field 3"])
    v = 1
    for row in range(rows):
        # Some have spaces, some not, to help test padding columns of different widths
        t.add_row([f"value {v*100}", f"value {v+1*100}", f"value {v+2*100}"])
        v += 3
    return t



t = helper_table(rows=3)
t.format = True
t.align = "r"
t.align["Field 1"] = "l"
t.valign["Field 1"] = "T"
t.add_autoindex("I")

print(t.get_string())
print(t.get_html_string())
```

The output of running this is shown below.

```
+------+-----------+-----------+-----------+
|  I   | Field 1   |   Field 2 |   Field 3 |
+------+-----------+-----------+-----------+
| 1000 | value 100 | value 101 | value 201 |
| 1001 | value 400 | value 104 | value 204 |
| 1002 | value 700 | value 107 | value 207 |
+------+-----------+-----------+-----------+
Traceback (most recent call last):
  File "tests/testy.py", line 24, in <module>
    print(t.get_html_string())
  File "/home/user/repo/prettytable/venv/lib/python3.8/site-packages/prettytable-3.2.1.dev44-py3.8.egg/prettytable/prettytable.py", line 2079, in get_html_string
  File "/home/user/repo/prettytable/venv/lib/python3.8/site-packages/prettytable-3.2.1.dev44-py3.8.egg/prettytable/prettytable.py", line 2199, in _get_formatted_html_string
TypeError: unhashable type: 'dict'
```

myheroyuki:
[...]

Issue #111, opened 2021-05-06
Alignment broken for default case


johnthagen:
I followed the README steps for doing column alignment:

```py
from prettytable import PrettyTable


def main() -> None:
    x = PrettyTable()
    x.align = "r"
    x.field_names = ["City name", "Area", "Population", "Annual Rainfall"]
    x.add_rows(
        [
            ["Adelaide", 1295, 1158259, 600.5],
            ["Brisbane", 5905, 1857594, 1146.4],
            ["Darwin", 112, 120900, 1714.7],
            ["Hobart", 1357, 205556, 619.5],
            ["Sydney", 2058, 4336374, 1214.8],
            ["Melbourne", 1566, 3806092, 646.9],
            ["Perth", 5386, 1554769, 869.4],
        ]
    )
    print(x)


if __name__ == "__main__":
    main()
```

But the output is not aligned:

```
+-----------+------+------------+-----------------+
| City name | Area | Population | Annual Rainfall |
+-----------+------+------------+-----------------+
|  Adelaide | 1295 |  1158259   |      600.5      |
|  Brisbane | 5905 |  1857594   |      1146.4     |
|   Darwin  | 112  |   120900   |      1714.7     |
|   Hobart  | 1357 |   205556   |      619.5      |
|   Sydney  | 2058 |  4336374   |      1214.8     |
| Melbourne | 1566 |  3806092   |      646.9      |
|   Perth   | 5386 |  1554769   |      869.4      |
+-----------+------+------------+-----------------+
```

Am I missing something?

- prettytable 2.1.0
- Python 3.9.1

johnthagen:
Shortly after reporting this I did some testing and realize that *when* `align` is set matters. This works:

```py
from prettytable import PrettyTable


def main() -> None:
    x = PrettyTable()

    x.field_names = ["City name", "Area", "Population", "Annual Rainfall"]
    x.add_rows(
        [
            ["Adelaide", 1295, 1158259, 600.5],
            ["Brisbane", 5905, 1857594, 1146.4],
            ["Darwin", 112, 120900, 1714.7],
            ["Hobart", 1357, 205556, 619.5],
            ["Sydney", 2058, 4336374, 1214.8],
            ["Melbourne", 1566, 3806092, 646.9],
            ["Perth", 5386, 1554769, 869.4],
        ]
    )
    x.align = "r"
    print(x)


if __name__ == "__main__":
    main()
```

[...]

Issue #102, opened 2021-03-30
Add header_align argument


rlabbe:
The idea here is to be able to align the header separately from the values. For example, when I print floats I like the decimal points to align, and I get that with align='r'. But then the header names are all the way to the right, and you end up with

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;x
&nbsp;&nbsp;&nbsp;&nbsp;12346.657

whereas I'd like:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;x
&nbsp;&nbsp;&nbsp;&nbsp;12346.657

hugovk:
I'd be open to looking at a PR for this.

hugovk:
Re-opening as https://github.com/jazzband/prettytable/pull/183 was reverted, a new PR is welcome, thanks!

myheroyuki:
Hi, I'm willing to give this one another attempt. @hugovk, what do you think the correct behavior should be. I think that `header_align` should take precedence if both `header_align` and `align` are set. It seems strange to me to allow the user to use `align` for a individual header fields when `header_align` exists, but I can see from the issue mentioned in the pull request it's needed to ensure backwards capability.

 I think the best solution is that `align` should work as always as long as no `header_align` is set. If the user ever does set `header align`, it will be used to set the alignment for header fields. This could be bit confusing though if `header_align` is set very early on, and the user later changes a specific header field's alignment at the end using `align`, as that align value would be ignored. I suppose that's fine as long as it's documented though.

hugovk:
Hi, thank you! Yes, that sounds about right :)

* If only `align` is set, it's the "global" option that applies to both header and body.

* If only `header_align` is set, use `header_align` for the header, no change to the body.

* If both `header_align` and `align` are set, use `header_align` for the header, and `align` for the body.

Ping also @osfrickler and @SeanMooney: how does this sound?

SeanMooney:
"""If only align is set, it's the "global" option that applies to both header and body."""

i think that would maintain the behaviour we are currently depending on in our test yes.

the rest also make logical sense to me.

PR #207, opened 2022-10-18
Alignment alteration

myheroyuki:
Here's a possible solution to a batch of issues with column alignment (#202, #199 , #102, #111). The root of these issues seemed to be the fact that the `align` attribute was filling two roles. It was a global alignment, set as a string, and also a per-column alignment, set as a dictionary. I fixed this by creating a new `Alignment` class that is basically a dictionary that contains a variable that stores the global alignment value.

The main reason I decided to break it out into it's own single class was that having it as a class allowed me to validate the attributes when the user sets it per column using square brackets. It still functions the same way for the user when they set it, update it, print it, or iterate over it, but it changes the return type to be an instance of `Alignment` rather than `dict`. 

I'd really like to get some comments on this one with some ideas on how I can improve it. I don't love the solution I came up with here, but other things I thought of either would change things from the user's perspective, not address all the issues, or make the code more confusing.


src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
     return width, height
 
 
+##############################
+# ALIGNMENT VALIDATORS       #
+##############################
+
+
+def validate_align(val):
```
hugovk: Should these three validators be private/internal and not public?

And let's add type hints.

```suggestion
def _validate_align(val: str) -> None:
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+    try:
+        assert val in ["l", "c", "r"]
```
```suggestion
        assert val in ("l", "c", "r")
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+    except AssertionError:
+        raise ValueError(f"Alignment {val} is invalid, use l, c or r")
+
+
+def validate_valign(val):
```
```suggestion
def _validate_valign(val: str | None) -> None:
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+    try:
+        assert val in ["t", "m", "b", None]
```
```suggestion
        assert val in ("t", "m", "b", None)
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+    except AssertionError:
+        raise ValueError(f"Alignment {val} is invalid, use t, m, b or None")
+
+
+def validate_header_align(val):
```
```suggestion
def _validate_header_align(val: str | None) -> None:
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+    try:
+        assert val in ["l", "c", "r", None]
```
```suggestion
        assert val in ("l", "c", "r", None)
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+##############################
+# ALIGNMENT VALIDATORS       #
+##############################
+
+
+def validate_align(val):
+    try:
+        assert val in ["l", "c", "r"]
+    except AssertionError:
+        raise ValueError(f"Alignment {val} is invalid, use l, c or r")
```
```suggestion
        raise ValueError(f"Alignment {val} is invalid: use l, c or r")
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+
+
+def validate_valign(val):
+    try:
+        assert val in ["t", "m", "b", None]
+    except AssertionError:
+        raise ValueError(f"Alignment {val} is invalid, use t, m, b or None")
```
```suggestion
        raise ValueError(f"Alignment {val} is invalid: use t, m, b or None")
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+
+
+def validate_header_align(val):
+    try:
+        assert val in ["l", "c", "r", None]
+    except AssertionError:
+        raise ValueError(f"Alignment {val} is invalid, use l, c, r or None")
```
```suggestion
        raise ValueError(f"Alignment {val} is invalid: use l, c, r or None")
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+
+
+# A dict extension to manage field alignment. Used to:
+# Have global alingment sepreate from field alignments.
```
```suggestion
# * separate global alignment from field alignments
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+# Validate values when setting for individual fields
```
```suggestion
# * validate values when setting individual fields
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+class Alignment(dict):
+    def __init__(self, align_type=ALIGN_TYPE_DEFAULT):
```
hugovk: Please add type hints to these methods

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+    try:
+        assert val in ["l", "c", "r", None]
+    except AssertionError:
+        raise ValueError(f"Alignment {val} is invalid, use l, c, r or None")
+
+
+# A dict extension to manage field alignment. Used to:
+# Have global alingment sepreate from field alignments.
+# Validate values when setting for individual fields
+class Alignment(dict):
```
hugovk: Also private/internal?
```suggestion
class _Alignment(dict):
```

src/prettytable/prettytable.py
```diff
@@ -74,6 +76,61 @@ def _get_size(text):
...
+            validate_valign(val)
+        elif self.align_type == ALIGN_TYPE_HEADER:
+            validate_header_align(val)
+        else:
+            validate_align(val)
+        dict.__setitem__(self, field_name, val)
+
+    def __repr__(self):
+        dictrepr = dict.__repr__(self)
+        return dictrepr
```
```suggestion
        return dict.__repr__(self)
```

Please also add a test case for this method.

src/prettytable/prettytable.py
```diff
@@ -1842,6 +1938,12 @@ def _stringify_header(self, options):
         for (field, width) in zip(self._field_names, self._widths):
             if options["fields"] and field not in options["fields"]:
                 continue
+            if field in self._header_align and self._header_align[field]:
+                align = self._header_align[field]
+            elif field in self._align:
+                align = self._align[field]
+            else:
+                align = self._align.alignment
```
hugovk: Please can you add a test case to cover this line?

tests/test_prettytable.py
```diff
@@ -384,6 +390,85 @@ def test_aligned_latex(
...
+        t.align = {}
+        t.header_align = {}
+
+        # Assert
+        for field in t.align:
+            assert t.align[field] == "c"
+        for field in t.valign:
+            assert t.valign[field] == "t"
+        for field in t.header_align:
+            assert t.header_align[field] == None
```
hugovk: Lint fix:
```suggestion
            assert t.header_align[field] is None
```

tests/test_prettytable.py
```diff
@@ -1437,6 +1523,41 @@ def test_style_align(self, style, expected):
...
+                id="MARKDOWN",
+            ),
+        ],
+    )
+    def test_style_header_align(self, style, expected):
+        # Arrange
+        t = helper_table()
+        t.field_names = ["L", "C", "R"]
+
+        assert t.header_align["L"] == None
```
hugovk: Lint fix:
```suggestion
        assert t.header_align["L"] is None
```

hugovk (commented):
Thanks for the PR! Here's some review suggestions.

PR #211, opened 2022-10-28
Add `py.typed` and update `setup.cfg`

phershbe:
As mentioned here: https://github.com/jazzband/prettytable/issues/203#issuecomment-1295330801

Created `py.typed` in `src/prettytable` as a marker file in order to indicate that the package supports typing.
Also added `Typing :: Typed` to `setup.cfg` as a `Trove classifier` in the `classifiers` list in order to help find and classify things on PyPI.


hugovk (approved):
Thanks!

PR #218, opened 2022-11-10
Updated pre-commit with mypy

phershbe:
Issue: https://github.com/jazzband/prettytable/issues/203

mypy is added to pre-commit but doesn't pass yet.  The error messages are below. I'm working on them but anybody else is welcome to pitch in as well.

```
src/prettytable/prettytable.py:137: error: Need type annotation for "_field_names" (hint: "_field_names: List[<type>] = ...")
src/prettytable/prettytable.py:138: error: Need type annotation for "_rows" (hint: "_rows: List[<type>] = ...")
src/prettytable/prettytable.py:150: error: Need type annotation for "_widths" (hint: "_widths: List[<type>] = ...")
src/prettytable/prettytable.py:209: error: Need type annotation for "_none_format" (hint: "_none_format: Dict[<type>, <type>] = ...")
src/prettytable/prettytable.py:2042: error: Argument 2 to "dumps" has incompatible type "**Dict[str, object]"; expected "bool"
src/prettytable/prettytable.py:2042: error: Argument 2 to "dumps" has incompatible type "**Dict[str, object]"; expected "Optional[Type[JSONEncoder]]"
src/prettytable/prettytable.py:2042: error: Argument 2 to "dumps" has incompatible type "**Dict[str, object]"; expected "Union[None, int, str]"
src/prettytable/prettytable.py:2042: error: Argument 2 to "dumps" has incompatible type "**Dict[str, object]"; expected "Optional[Tuple[str, str]]"
src/prettytable/prettytable.py:2042: error: Argument 2 to "dumps" has incompatible type "**Dict[str, object]"; expected "Optional[Callable[[Any], Any]]"
src/prettytable/prettytable.py:2423: error: Need type annotation for "tables" (hint: "tables: List[<type>] = ...")
src/prettytable/prettytable.py:2424: error: Need type annotation for "last_row" (hint: "last_row: List[<type>] = ...")
src/prettytable/prettytable.py:2425: error: Need type annotation for "rows" (hint: "rows: List[<type>] = ...")
src/prettytable/__init__.py:52: error: Name "importlib_metadata" already defined (by an import)
src/prettytable/colortable.py:37: error: Self argument missing for a non-static method (or an invalid type for self)
tests/test_prettytable.py:304: error: Module "prettytable" is not valid as a type
tests/test_prettytable.py:305: error: prettytable? has no attribute "get_string"
tests/test_prettytable.py:309: error: Module "prettytable" is not valid as a type
tests/test_prettytable.py:310: error: prettytable? has no attribute "get_html_string"
tests/test_prettytable.py:314: error: Module "prettytable" is not valid as a type
tests/test_prettytable.py:315: error: prettytable? has no attribute "get_latex_string"
tests/test_prettytable.py:319: error: Module "prettytable" is not valid as a type
tests/test_prettytable.py:320: error: prettytable? has no attribute "field_names"
tests/test_prettytable.py:328: error: prettytable? has no attribute "get_string"
tests/test_prettytable.py:367: error: Module "prettytable" is not valid as a type
tests/test_prettytable.py:369: error: prettytable? has no attribute "get_string"
tests/test_prettytable.py:370: error: prettytable? has no attribute "get_string"
tests/test_prettytable.py:374: error: Module "prettytable" is not valid as a type
tests/test_prettytable.py:376: error: prettytable? has no attribute "get_html_string"
tests/test_prettytable.py:377: error: prettytable? has no attribute "get_html_string"
tests/test_prettytable.py:381: error: Module "prettytable" is not valid as a type
tests/test_prettytable.py:383: error: prettytable? has no attribute "get_latex_string"
tests/test_prettytable.py:384: error: prettytable? has no attribute "get_latex_string"
tests/test_prettytable.py:420: error: Module "prettytable" is not valid as a type
tests/test_prettytable.py:421: error: prettytable? has no attribute "get_string"
tests/test_prettytable.py:422: error: prettytable? has no attribute "get_string"
tests/test_prettytable.py:528: error: Module "prettytable" is not valid as a type
tests/test_prettytable.py:529: error: prettytable? has no attribute "get_string"
tests/test_prettytable.py:533: error: Module "prettytable" is not valid as a type
tests/test_prettytable.py:534: error: prettytable? has no attribute "get_string"
tests/test_prettytable.py:537: error: Incompatible types in assignment (expression has type "Set[int]", variable has type "List[int]")
tests/test_prettytable.py:784: error: "PrettyTable" has no attribute "caching"
Found 41 errors in 4 files (checked 5 source files)
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,5 +57,10 @@ repos:
         args: [--prose-wrap=always, --print-width=88]
         exclude: ^.github/ISSUE_TEMPLATE/bug_report.md$
 
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: "v0.982"
+    hooks:
+      - id: mypy
+
 ci:
   autoupdate_schedule: quarterly
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,6 +61,8 @@ repos:
     rev: "v0.982"
     hooks:
       - id: mypy
+        args: [--pretty, --show-error-codes]
+        exclude: ^tests/
 
 ci:
   autoupdate_schedule: quarterly
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,6 +61,7 @@ repos:
     rev: "v0.982"
     hooks:
       - id: mypy
+        additional_dependencies: [types-colorama, types-setuptools]
         args: [--pretty, --show-error-codes]
         exclude: ^tests/
 
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -45,7 +45,7 @@ from html import escape
 from html.parser import HTMLParser
 from typing import Any
 
-import wcwidth
+import wcwidth  # type: ignore
 
 # hrule styles
 FRAME = 0
@@ -134,8 +134,8 @@ class PrettyTable:
         self.encoding = kwargs.get("encoding", "UTF-8")
 
         # Data
-        self._field_names = []
-        self._rows = []
+        self._field_names: list[str] = []
+        self._rows: list[list] = []
         self.align = {}
         self.valign = {}
         self.max_width = {}
@@ -147,7 +147,7 @@ class PrettyTable:
         if field_names:
             self.field_names = field_names
         else:
-            self._widths = []
+            self._widths: list[int] = []
 
         # Options
         self._options = [
@@ -206,7 +206,7 @@ class PrettyTable:
         self._start = kwargs["start"] or 0
         self._end = kwargs["end"] or None
         self._fields = kwargs["fields"] or None
-        self._none_format = {}
+        self._none_format: dict[None, None] = {}
 
         if kwargs["header"] in (True, False):
             self._header = kwargs["header"]
@@ -2031,7 +2031,7 @@ class PrettyTable:
         """
 
         options = self._get_options(kwargs)
-        json_options = dict(indent=4, separators=(",", ": "), sort_keys=True)
+        json_options: Any = dict(indent=4, separators=(",", ": "), sort_keys=True)
         json_options.update(
             {key: value for key, value in kwargs.items() if key not in options}
         )
@@ -2423,9 +2423,9 @@ class TableHandler(HTMLParser):
     def __init__(self, **kwargs) -> None:
         HTMLParser.__init__(self)
         self.kwargs = kwargs
-        self.tables = []
-        self.last_row = []
-        self.rows = []
+        self.tables: list[list] = []
+        self.last_row: list[str] = []
+        self.rows: list[Any] = []
         self.max_row_width = 0
         self.active = None
         self.last_content = ""
--- a/src/prettytable/__init__.py
+++ b/src/prettytable/__init__.py
@@ -49,6 +49,6 @@ try:
     import importlib.metadata as importlib_metadata
 except ImportError:
     # <Python 3.7 and lower
-    import importlib_metadata
+    import importlib_metadata  # type: ignore
 
 __version__ = importlib_metadata.version(__name__)
--- a/src/prettytable/colortable.py
+++ b/src/prettytable/colortable.py
@@ -4,13 +4,11 @@ from .prettytable import PrettyTable
 
 try:
     from colorama import init
+
+    init()
 except ImportError:
-    # Do nothing if not installed
-    def init():
-        pass
+    pass
 
-
-init()
 
 RESET_CODE = "\x1b[0m"
 
@@ -34,6 +32,7 @@ class Theme:
         self.junction_char = junction_char
         self.junction_color = Theme.format_code(junction_color)
 
+    @staticmethod
     def format_code(s: str) -> str:
         """Takes string and intelligently puts it into an ANSI escape sequence"""
         if s.strip() == "":


phershbe:
@hugovk I started to make some type annotations and tried to upload them but Git wouldn't accept the commit with mypy not passing the pre-commit. I'm going to finish making them and try adding them tomorrow. I'm a little bit in over my head here but I'm starting to get it, I'll need it checked obviously but I'll also include questions on the things that I'm most uncertain about.

.pre-commit-config.yaml
```diff
@@ -57,5 +57,10 @@ repos:
         args: [--prose-wrap=always, --print-width=88]
         exclude: ^.github/ISSUE_TEMPLATE/bug_report.md$
 
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: "v0.982"
+    hooks:
+      - id: mypy
```
hugovk, line 63: Let's add these arguments to make the errors a bit more helpful, and also exclude the tests to narrow the scope:
```suggestion
      - id: mypy
        args: [--pretty, --show-error-codes]
        exclude: ^tests/
```

hugovk:
Sure, let's tackle them bit by bit!

---

If you have pre-commit installed locally, and the commit fails because mypy isn't yet happy, you can bypass it using the `-n` flag:

```
-n, --no-verify       bypass pre-commit and commit-msg hooks
```

For example:

```sh
git commit -m "My commit message" -n
```

---


For this:
```
src/prettytable/__init__.py:52: error: Cannot find implementation or library
stub for module named "importlib_metadata"  [import]
        import importlib_metadata
    ^
src/prettytable/__init__.py:52: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
src/prettytable/__init__.py:52: error: Name "importlib_metadata" already
defined (by an import)  [no-redef]
        import importlib_metadata
        ^
```

That's coming from here:

```python
try:
    # Python 3.8+
    import importlib.metadata as importlib_metadata
except ImportError:
    # <Python 3.7 and lower
    import importlib_metadata
```

It's getting a bit confused by the multiple imports. In real life, we can only ever import one of them, but I don't think mypy can handle this. So the usual thing to do is just tell mypy to ignore the second one:

```diff
 try:
     import importlib.metadata as importlib_metadata
 except ImportError:
     # <Python 3.7 and lower
-    import importlib_metadata
+    import importlib_metadata  # type: ignore
```

.pre-commit-config.yaml
```diff
@@ -57,5 +57,12 @@ repos:
...
```
hugovk, line 63: For these:

```
setup.py:3: error: Skipping analyzing "setuptools": module is installed, but
missing library stubs or py.typed marker  [import]
    from setuptools import setup

src/prettytable/colortable.py:6: error: Cannot find implementation or library
stub for module named "colorama"  [import]
        from colorama import init
```

This is because we're importing the setuptools and Colorama modules but they don't have type hints (their tracking issues: https://github.com/pypa/setuptools/issues/2345, https://github.com/tartley/colorama/issues/206).

However, just like PrettyTable didn't type hints before https://github.com/jazzband/prettytable/pull/205, there are stubs published by Typeshed and available at https://pypi.org/project/types-colorama/ and https://pypi.org/project/types-setuptools/

So we can make use of those by telling pre-commit to install it like this:

```suggestion
      - id: mypy
        additional_dependencies: [types-colorama, types-setuptools]
```

(If running mypy outside of pre-commit, `pip install types-colorama types-setuptools`)

phershbe:
@hugovk I'm working on it now. I said that I would submit it yesterday but I got delayed because I couldn't figure out why my `git clone` wasn't showing the changes in the `pre-commit` file. I was on the wrong branch, and finally got it figured out with `git checkout`. I'll have it in soon.

hugovk:
All good! As before, absolutely no rush with this, let's take our time :)

This one:
```
src/prettytable/colortable.py:37: error: Self argument missing for a non-static
method (or an invalid type for self)  [misc]
        def format_code(s: str) -> str:
```

Looks like we've found a bug :)  There is no `self` for that method and it doesn't modify anything from the class, so let's make it static:

```diff
diff --git a/src/prettytable/colortable.py b/src/prettytable/colortable.py
index dd5495a..625c89e 100644
--- a/src/prettytable/colortable.py
+++ b/src/prettytable/colortable.py
@@ -34,6 +34,7 @@ class Theme:
         self.junction_char = junction_char
         self.junction_color = Theme.format_code(junction_color)
 
+    @staticmethod
     def format_code(s: str) -> str:
         """Takes string and intelligently puts it into an ANSI escape sequence"""
         if s.strip() == "":
```

hugovk:
For this:

```
src/prettytable/colortable.py:9: error: All conditional function variants must
have identical signatures  [misc]
        def init():
        ^
```

The code in question is:

```python
try:
    from colorama import init
except ImportError:
    # Do nothing if not installed
    def init():
        pass


init()
```

Colorama is used to set up the terminal for printing colour, mainly needed for Windows where it can otherwise be a bit tricky.

But we don't require Colorama as a dependency, and so only call `init` if it happens to be installed as a dependency.

But let's not have to worry about keeping in sync the signatures of the real `init` with our fake one, and instead refactor it so remove our dummy `init`, something like:

```diff
diff --git a/src/prettytable/colortable.py b/src/prettytable/colortable.py
index dd5495a..e1224ea 100644
--- a/src/prettytable/colortable.py
+++ b/src/prettytable/colortable.py
@@ -4,13 +4,12 @@ from .prettytable import PrettyTable
 
 try:
     from colorama import init
+
+    init()
 except ImportError:
     # Do nothing if not installed
-    def init():
-        pass
-
+    pass
 
-init()
 
 RESET_CODE = "\x1b[0m"
 
```

hugovk:
This one is like setuptools and Colorama, no type hints in the module:

```
src/prettytable/prettytable.py:48: error: Cannot find implementation or library
stub for module named "wcwidth"  [import]
    import wcwidth
```

Except unfortunately there are no type hints on Typeshed either. I think we can get away with just ignoring this one:

```diff
diff --git a/src/prettytable/prettytable.py b/src/prettytable/prettytable.py
index 18379ce..29fe714 100644
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -45,7 +45,7 @@ from html import escape
 from html.parser import HTMLParser
 from typing import Any
 
-import wcwidth
+import wcwidth  # type: ignore
 
 # hrule styles
 FRAME = 0
```

phershbe:
@hugovk Okay, I finally finished. I got the static method correction by myself but was super stuck on a few others and then realized that you had made notes about them above, awesome. I'm not super confident about types that I added, they make mypy pass so I guess that's a good sign, but you might want to check them.

src/prettytable/prettytable.py
```diff
@@ -2423,9 +2423,9 @@ class TableHandler(HTMLParser):
     def __init__(self, **kwargs) -> None:
         HTMLParser.__init__(self)
         self.kwargs = kwargs
-        self.tables = []
-        self.last_row = []
-        self.rows = []
+        self.tables: list[list] = []
+        self.last_row: list[str] = []
+        self.rows: list[Any] = []
```
hugovk, line 2428: It's good to avoid `Any` if possible, as it's like having no typing.

So `self.rows` is a list of tuples:

```suggestion
        self.rows: list[tuple] = []
```

But it's also good practice to give more accurate types for things like tuples, lists and dicts.

The next step after this PR would be to add the `--strict` flag to mypy, which will complain about these, but then there are over 300 errors! Anyway, we can maybe do a few of these here. (And there's an automated tool that can help with many of those, but that can also wait.)

So this tuple is a list (of table data) and then a Boolean:

```suggestion
        self.rows: list[tuple[list, bool]] = []
```

Ideally we should also say what can be in the list. This can be things like strings, ints, floats, I guess Booleans too, but maybe others? Maybe we can worry about enumerating that later (and when we do we can use a [_type alias_](https://docs.python.org/3/library/typing.html#type-aliases), like a variable for types).

So I think `list[tuple[list, bool]]` is enough for here :)

src/prettytable/prettytable.py
```diff
@@ -2423,9 +2423,9 @@ class TableHandler(HTMLParser):
     def __init__(self, **kwargs) -> None:
         HTMLParser.__init__(self)
         self.kwargs = kwargs
-        self.tables = []
-        self.last_row = []
-        self.rows = []
+        self.tables: list[list] = []
```
hugovk, line 2426: `list` of `PrettyTable`s?
```suggestion
        self.tables: list[PrettyTable] = []
```

src/prettytable/prettytable.py
```diff
@@ -2031,7 +2031,7 @@ def get_json_string(self, **kwargs) -> str:
         """
 
         options = self._get_options(kwargs)
-        json_options = dict(indent=4, separators=(",", ": "), sort_keys=True)
+        json_options: Any = dict(indent=4, separators=(",", ": "), sort_keys=True)
```
hugovk, line 2034: At least a dict:

```suggestion
        json_options: dict = dict(indent=4, separators=(",", ": "), sort_keys=True)
```

Ideally we should list the types of the keys and values, but that can wait for `--strict`.

src/prettytable/prettytable.py
```diff
@@ -206,7 +206,7 @@ def __init__(self, field_names=None, **kwargs) -> None:
         self._start = kwargs["start"] or 0
         self._end = kwargs["end"] or None
         self._fields = kwargs["fields"] or None
-        self._none_format = {}
+        self._none_format: dict[None, None] = {}
```
hugovk, line 209: This can at least also have a string for keys and values:

```suggestion
        self._none_format: dict[str | None, str | None] = {}
```

hugovk (commented):
Good stuff! Just a few little comments and this is good for merge.

src/prettytable/prettytable.py
phershbe: @hugovk This is all super useful, I can't thank you enough for your personal feedback!

phershbe: @hugovk Okay, I'll read about the `--strict` flag, thank you.

hugovk: > @hugovk This is all super useful, I can't thank you enough for your personal feedback!

Happy to help, I'm also learning stuff!

> @hugovk Okay, I'll read about the `--strict` flag, thank you.

It might also be that we don't want to enable `--strict` on the CI and fix all of those warnings, maybe just a few. But one step at a time, and we can do some with https://github.com/JelleZijlstra/autotyping :)

phershbe:
@hugovk It seems that I had forgotten to change this from a draft pull request to a pull request, sorry about my confusion. It should be ready now!

hugovk:
Thank you, merged! 🚀

PR #216, opened 2022-11-08
Widen tables with long titles

etjones:
Recalculated this to scale the table correctly.

- Added `test_prettytable.test_all_lengths_equal_with_long_title()` to capture this this error as well.

**Original behavior:**
```

Without title:
+---------+---------+
| field a | field b |
+---------+---------+
|    a1   |    b1   |
|    a2   |    b2   |
+---------+---------+

With title:
+-----------------------------------------------------------------+
| === 80 char str ================================================================ |
+--------------------------------+--------------------------------+
|            field a             |            field b             |
+--------------------------------+--------------------------------+
|               a1               |               b1               |
|               a2               |               b2               |
+--------------------------------+--------------------------------+
```

**New behavior:**
``` 
With title:
+-----------------------------------------------------------------------------------+
|  === 80 char str ================================================================ |
+-----------------------------------------+-----------------------------------------+
|                 field a                 |                 field b                 |
+-----------------------------------------+-----------------------------------------+
|                    a1                   |                    b1                   |
|                    a2                   |                    b2                   |
+-----------------------------------------+-----------------------------------------+
```

PR #219, opened 2022-11-15
Fix to `min_table_width` parameter

kzwolenik95:
Before the change specifying **min_table_width** parameter changed the width of the table but the width value seemed random.

After the change width is exactly the parameter value when the content is not wide enough. At the same time this pull request fixes / preserves behavior from #216 (Long title doesn't break table layout)

Added tests for odd and even **min_table_width** for one, two, three columns tables.

min_table_width=50
```
+--------------+------------------+--------------+
|      F1      |        F2        |      F3      |
+--------------+------------------+--------------+
|     v 1      |       None       |     v 2      |
+--------------+------------------+--------------+
```

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1583,16 +1583,18 @@ class PrettyTable:
                 title_width = 0
             min_table_width = self.min_table_width or 0
             min_width = max(title_width, min_table_width)
-            if options['border']:
+            if options["border"]:
                 borders = len(widths) + 1
             else:
-                if options['preserve_internal_border']:
+                if options["preserve_internal_border"]:
                     borders = len(widths) - 1
                 else:
                     borders = 0
 
             # substract padding for each column and borders
-            min_width -= (sum([sum(self._get_padding_widths(options)) for _ in widths]) + borders)
+            min_width -= (
+                sum([sum(self._get_padding_widths(options)) for _ in widths]) + borders
+            )
             # what is being scaled is content so we sum column widths
             content_width = sum(widths) or 1
 
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1583,14 +1583,25 @@ class PrettyTable:
                 title_width = 0
             min_table_width = self.min_table_width or 0
             min_width = max(title_width, min_table_width)
-            table_width = self._compute_table_width(options)
-            if table_width < min_width:
-                # To make enough space for the title, all columns have to
-                # grow enough that their combined widths (and the borders
-                # between them) are equal to the length of the title.
-                border_char_count = 3 * (len(widths) - 1)
-                scale = 1.0 * (len(options["title"]) - border_char_count) / sum(widths)
-                widths = [int(math.ceil(w * scale)) for w in widths]
+            if options['border']:
+                borders = len(widths) + 1
+            else:
+                if options['preserve_internal_border']:
+                    borders = len(widths) - 1
+                else:
+                    borders = 0
+
+            # substract padding for each column and borders
+            min_width -= (sum([sum(self._get_padding_widths(options)) for _ in widths]) + borders)
+            # what is being scaled is content so we sum column widths
+            content_width = sum(widths) or 1
+
+            if content_width < min_width:
+                # Grow widths in proportion
+                scale = 1.0 * min_width / content_width
+                widths = [int(math.floor(w * scale)) for w in widths]
+                if sum(widths) < min_width:
+                    widths[-1] += min_width - sum(widths)
                 self._widths = widths
 
     def _get_padding_widths(self, options):
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1587,13 +1587,13 @@ class PrettyTable:
                 borders = len(widths) + 1
             else:
                 if options['preserve_internal_border']:
-                    borders = len(widths) - 1
+                    borders = len(widths)
                 else:
                     borders = 0
 
-            # substract padding for each column and borders
+            # Subtract padding for each column and borders
             min_width -= (sum([sum(self._get_padding_widths(options)) for _ in widths]) + borders)
-            # what is being scaled is content so we sum column widths
+            # What is being scaled is content so we sum column widths
             content_width = sum(widths) or 1
 
             if content_width < min_width:
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1585,11 +1585,10 @@ class PrettyTable:
             min_width = max(title_width, min_table_width)
             if options["border"]:
                 borders = len(widths) + 1
+            elif options["preserve_internal_border"]:
+                borders = len(widths)
             else:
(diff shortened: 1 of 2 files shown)


src/prettytable/prettytable.py
```diff
@@ -1583,14 +1583,27 @@ def _compute_widths(self, rows, options):
...
-                # grow enough that their combined widths (and the borders
-                # between them) are equal to the length of the title.
-                border_char_count = 3 * (len(widths) - 1)
-                scale = 1.0 * (len(options["title"]) - border_char_count) / sum(widths)
-                widths = [int(math.ceil(w * scale)) for w in widths]
+            if options["border"]:
+                borders = len(widths) + 1
+            else:
+                if options["preserve_internal_border"]:
+                    borders = len(widths) - 1
```
hugovk: Please could you add tests to cover this line?

src/prettytable/prettytable.py
```diff
@@ -1583,14 +1583,27 @@ def _compute_widths(self, rows, options):
...
+                else:
+                    borders = 0
```
hugovk: And to cover this line?

src/prettytable/prettytable.py
```diff
@@ -1583,14 +1583,27 @@ def _compute_widths(self, rows, options):
...
+
+            # substract padding for each column and borders
```
```suggestion
            # Subtract padding for each column and borders
```

src/prettytable/prettytable.py
```diff
@@ -1583,14 +1583,27 @@ def _compute_widths(self, rows, options):
...
+            min_width -= (
+                sum([sum(self._get_padding_widths(options)) for _ in widths]) + borders
+            )
+            # what is being scaled is content so we sum column widths
```
```suggestion
            # What is being scaled is content so we sum column widths
```

tests/test_prettytable.py
```diff
@@ -1912,6 +1912,106 @@ def test_jupyter_repr(self, row_prettytable: PrettyTable):
         assert row_prettytable._repr_html_() == row_prettytable.get_html_string()
 
 
+class TestMinTableWidth:
+    def test_even_min_table_width_one_column(self):
+        for i in range(15):
+            x = prettytable.PrettyTable()
+            x.field_names = ["Test table"]
+            x.add_row(["X" * i])
```
hugovk: Please could you use a more descriptive variable name for this? For example:
```suggestion
        for col_width in range(15):
            x = prettytable.PrettyTable()
            x.field_names = ["Test table"]
            x.add_row(["X" * col_width])
```

And below.

tests/test_prettytable.py
```diff
@@ -1912,6 +1912,106 @@ def test_jupyter_repr(self, row_prettytable: PrettyTable):
...
+            desired_width = 20
+            x.min_table_width = desired_width
+            t = x.get_string()
+            print(t)
```
hugovk: Let's remove `print`s.
```suggestion
```

tests/test_prettytable.py
```diff
@@ -1912,6 +1912,106 @@ def test_jupyter_repr(self, row_prettytable: PrettyTable):
         assert row_prettytable._repr_html_() == row_prettytable.get_html_string()
 
 
+class TestMinTableWidth:
+    def test_even_min_table_width_one_column(self):
```
hugovk: These tests all look pretty similar, can we use `pytest.mark.parametrize`?

hugovk (commented):
Thank you for testing PrettyTable before release, it's really good to catch these things early 👍 

@etjones Please could you review?

src/prettytable/prettytable.py
kzwolenik95: I have added tests to tables with only internal borders, I found out that it have to be `borders = len(widths)` actually, after this change all tests passed
When table is rendered with only internal borders, horizontal line is one character shorter, I have taken this into the account when writing test but I don't know if this is the correct behavior. Part of the code I have touched doesn't affect borders

src/prettytable/prettytable.py
kzwolenik95: I included tables without borders in the tests, all passed correctly

src/prettytable/prettytable.py
kzwolenik95: done

src/prettytable/prettytable.py
kzwolenik95: done

tests/test_prettytable.py
kzwolenik95: I used your suggestion

tests/test_prettytable.py
kzwolenik95: Done

tests/test_prettytable.py
kzwolenik95: I have parametrized them using `pytest.mark.parametrize`
Thanks for suggestion

etjones:
Nice work, Krzysztof. I rushed that PR without doing as much testing as I should have. I'm grateful to you for cleaning up my mess.

tests/test_prettytable.py
```diff
@@ -1912,6 +1912,59 @@ def test_jupyter_repr(self, row_prettytable: PrettyTable):
...
+            x = prettytable.PrettyTable()
+            x.border = border
+            x.preserve_internal_border = internal_border
+            x.field_names = fields
+            x.add_row(["X" * col_width] + ["" for _ in range(len(fields) - 1)])
+            x.min_table_width = desired_width
+            t = x.get_string()
+            if border == False and internal_border == False:
+                assert [len(x) for x in t.split("\n")] == [desired_width, desired_width]
+            elif border == False and internal_border == True:
```
hugovk: To appease the linter:

```
tests/test_prettytable.py:1950:23: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
tests/test_prettytable.py:1950:52: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
tests/test_prettytable.py:1952:25: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
tests/test_prettytable.py:1952:54: E712 comparison to True should be 'if cond is True:' or 'if cond:'
```

```suggestion
            if border is False and internal_border is False:
                assert [len(x) for x in t.split("\n")] == [desired_width, desired_width]
            elif border is False and internal_border is True:
```

src/prettytable/prettytable.py
```diff
@@ -1583,14 +1583,27 @@ def _compute_widths(self, rows, options):
...
-                border_char_count = 3 * (len(widths) - 1)
-                scale = 1.0 * (len(options["title"]) - border_char_count) / sum(widths)
-                widths = [int(math.ceil(w * scale)) for w in widths]
+            if options["border"]:
+                borders = len(widths) + 1
+            else:
+                if options["preserve_internal_border"]:
+                    borders = len(widths)
+                else:
+                    borders = 0
```
hugovk: Can we simplify this?

```suggestion
            if options["border"]:
                borders = len(widths) + 1
            elif options["preserve_internal_border"]:
                borders = len(widths)
            else:
                borders = 0
```

hugovk:
Thanks!

PR #223, opened 2022-12-06
Refactor old %-formatting to use format() builtin

mwtoews:
The old `%`-formatting is not ageing well in modern Python code, and has largely been superseded by [format strings](https://docs.python.org/3/library/string.html#format-string-syntax) driven by the [`format()`](https://docs.python.org/3/library/functions.html#format) built-in. This PR pivots away from the `%` operator when used for string operations.

One notable change is to the behaviour of `int_format` and `float_format`, which have been changed to optionally allowed to end with alpha char, e.g. `#08x` or `8.3G`. Previously, only `float_format` optionally ended with `f`.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -488,9 +488,9 @@ class PrettyTable:
         try:
             assert isinstance(val, str)
             if not val[-1].isalpha():
-                val = f"{val}d"
+                val += "d"
             format(1, val)  # format an example int
-        except AssertionError:
+        except (AssertionError, ValueError):
             raise ValueError(
                 f"Invalid value for {name}. Must be an integer format string."
             )
@@ -501,7 +501,7 @@ class PrettyTable:
         try:
             assert isinstance(val, str)
             if not val[-1].isalpha():
-                val = f"{val}f"
+                val += "f"
             format(1.2, val)  # format an example float
         except (AssertionError, ValueError):
             raise ValueError(
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -487,7 +487,9 @@ class PrettyTable:
             return
         try:
             assert isinstance(val, str)
-            assert val.isdigit()
+            if not val[-1].isalpha():
+                val = f"{val}d"
+            format(1, val)  # format an example int
         except AssertionError:
             raise ValueError(
                 f"Invalid value for {name}. Must be an integer format string."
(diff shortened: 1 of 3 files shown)


src/prettytable/prettytable.py
```diff
@@ -1522,9 +1518,15 @@ def copy(self):
 
     def _format_value(self, field, value):
         if isinstance(value, int) and field in self._int_format:
-            return ("%%%sd" % self._int_format[field]) % value
+            int_format = self._int_format.get(field, "")
+            if not int_format[-1:].isalpha():
+                int_format += "d"
```
hugovk, line 1523: Any particular reason to use f-strings in the validate functions and concatenation here?

hugovk (commented):
Thanks for this! Please could you add/adjust tests to cover the new alpha char endings?

src/prettytable/prettytable.py
mwtoews: I wasn't certain that references to str variables would change in a function, but these are immutable so it's fine to use `+=` for both instances.

mwtoews:
I've added a few different tests for a few different format codes. Not sure if something else was expected, so please note and I can pivot the last commit.

src/prettytable/prettytable.py
```diff
@@ -487,8 +487,10 @@ def _validate_int_format(self, name, val):
             return
         try:
             assert isinstance(val, str)
-            assert val.isdigit()
-        except AssertionError:
+            if not val[-1].isalpha():
+                val += "d"
+            format(1, val)  # format an example int
+        except (AssertionError, ValueError):
```
hugovk, line 493: Tests look good, please could you also add some cases to cover this invalid branch?

src/prettytable/prettytable.py
```diff
@@ -498,16 +500,10 @@ def _validate_float_format(self, name, val):
...
-            assert (
-                bits[1] == ""
-                or bits[1].isdigit()
-                or (bits[1][-1] == "f" and bits[1].rstrip("f").isdigit())
-            )
-        except AssertionError:
+            if not val[-1].isalpha():
+                val += "f"
+            format(1.2, val)  # format an example float
+        except (AssertionError, ValueError):
```
hugovk, line 506: And this too? Thanks!

mwtoews:
Additional tests for invalid formats are added.

Note that using `format()` is a bit more strict than the old % formatters, and will require all values of the column to be int or float types. E.g., consider:
```python
"%3d" % 24.6  # ' 24'
format(24.6, "3d")  # ValueError: Unknown format code 'd' for object of type 'float'
```
this could have some implications for end-users that have mixed types in a column with either format code.

hugovk:
That could be a problem, we don't really want to introduce breaking changes here.

mwtoews:
I'd hate to break things. One solution is to catch an exception, show a warning about mixing datatypes and fallback with `%`-formatting. Is there an example table that would have mixed types? Also, how is missing data encoded?

_Update_: I've been trying to make a breaking example, but I can't seem to create one. A field can have mixed types, and a definition for each `int_format` and `float_format`, and everything seems to work as expected. And I've found missing cells are None (as expected, formatted by `none_format`). If a breaking example is found, I have a pending commit to fall-back to `int_format.__mod__(value)` / `float_format.__mod__(value)`.

PR #241, opened 2023-05-06
Add `get_formatted_string()` convenience function

rickwporter:
In a couple uses of `PrettyTable`, I have added a similar function to allow for less code from my CLI projects. In those projects, I allow specifying the output format type as an arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,8 +205,8 @@ your table to a file or insert it into a GUI.
 
 The table can be displayed in several different formats using `get_formatted_string`
 just by changing the `out_format=<text|html|json|csv|latex>`. This function passes
-through arguments to the functions that render the table, so additional arguments
-can be given. The provides an easy means to let a user choose the output formatting.
+through arguments to the functions that render the table, so additional arguments can be
+given. The provides an easy means to let a user choose the output formatting.
 
 ```python
 def my_cli_function(table_format: str = 'text'):
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ your table to a file or insert it into a GUI.
 The table can be displayed in several different formats using `get_formatted_string` by
 changing the `out_format=<text|html|json|csv|latex>`. This function passes through
 arguments to the functions that render the table, so additional arguments can be given.
-The provides an easy means to let a user choose the output formatting.
+This provides a way to let a user choose the output formatting.
 
 ```python
 def my_cli_function(table_format: str = 'text'):
--- a/README.md
+++ b/README.md
@@ -203,6 +203,17 @@ This string is guaranteed to look exactly the same as what would be printed by d
 `print(x)`. You can now do all the usual things you can do with a string, like write
 your table to a file or insert it into a GUI.
 
+The table can be displayed in several different formats using `get_formatted_string`
+just by changing the `out_format=<text|html|json|csv|latex>`. This function passes
+through arguments to the functions that render the table, so additional arguments
+can be given. The provides an easy means to let a user choose the output formatting.
+
+```python
+def my_cli_function(table_format: str = 'text'):
+  ...
+  print(x.get_formatted_string(table_format))
+```
+
 #### Controlling which data gets displayed
 
 If you like, you can restrict the output of `print(x)` or `x.get_string` to only the
--- a/README.md
+++ b/README.md
@@ -203,10 +203,10 @@ This string is guaranteed to look exactly the same as what would be printed by d
 `print(x)`. You can now do all the usual things you can do with a string, like write
 your table to a file or insert it into a GUI.
 
-The table can be displayed in several different formats using `get_formatted_string`
-by changing the `out_format=<text|html|json|csv|latex>`. This function passes
-through arguments to the functions that render the table, so additional arguments can be
-given. The provides an easy means to let a user choose the output formatting.
+The table can be displayed in several different formats using `get_formatted_string` by
+changing the `out_format=<text|html|json|csv|latex>`. This function passes through
+arguments to the functions that render the table, so additional arguments can be given.
+The provides an easy means to let a user choose the output formatting.
 
 ```python
 def my_cli_function(table_format: str = 'text'):
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ This string is guaranteed to look exactly the same as what would be printed by d
 your table to a file or insert it into a GUI.
 
 The table can be displayed in several different formats using `get_formatted_string`
-just by changing the `out_format=<text|html|json|csv|latex>`. This function passes
+by changing the `out_format=<text|html|json|csv|latex>`. This function passes
 through arguments to the functions that render the table, so additional arguments can be
 given. The provides an easy means to let a user choose the output formatting.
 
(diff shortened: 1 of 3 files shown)


src/prettytable/prettytable.py
```diff
@@ -1511,6 +1511,27 @@ def clear(self) -> None:
     def copy(self):
         return copy.deepcopy(self)
 
+    def get_output(self, format: str = "text", **kwargs) -> str:
```
hugovk: Kind of a pity `get_string` is already used for the plaintext version :)

Please could you rename `format` to something that doesn't shadow the builtin?

hugovk (changes requested):

src/prettytable/prettytable.py
rickwporter: Thanks for the rename suggestion -- I sometimes forget the builtins, since most editors do NOT warn you about them.

I think it would work if I renamed `get_string()` to `get_text_string()` and then `get_output()` to `get_string()`. Should I try that change out?

adilhussain540: Based on what the function is doing, I'd suggest renaming it to `get_formatted_output`.

src/prettytable/prettytable.py
```diff
@@ -1511,6 +1511,28 @@ def clear(self) -> None:
...
+        if out_format == "json":
+            return self.get_json_string(**kwargs)
+        if out_format == "csv":
+            return self.get_csv_string(**kwargs)
+        if out_format == "latex":
+            return self.get_latex_string(**kwargs)
+        raise ValueError(
+            f"Invalid format {out_format}. "
+            "Must be one of: text, html, json, csv, or latex"
+        )
```
adilhussain540, line 1534:
```suggestion
        supported_formats = {
            "text": self.get_string,
            "html": self.get_html_string,
            "json": self.get_json_string,
            "csv": self.get_csv_string,
            "latex": self.get_latex_string,
        }
        if out_format in supported_formats:
            return supported_formats[out_format](**kwargs)
        
        raise ValueError(
            f"Invalid format {out_format}. "
            f"Must be one of: {', '.join(supported_formats)}"
        )
```

Let's try to minimize code repetition.

adilhussain540 (changes requested):

src/prettytable/prettytable.py
hugovk: I think the original is clearer in this case.

src/prettytable/prettytable.py
rickwporter: Would `get_formatted_string` be the better name, since it is calling a `get_*string` method?

src/prettytable/prettytable.py
adilhussain540: @hugovk Yeah I agree, looks shorter but less readable now.
Thoughts on the latest version?

src/prettytable/prettytable.py
adilhussain540: sounds good to me

rickwporter: Here's a diff (on top of my PR) where I rename `get_string` to `get_text_string` and `get_output` to `get_string`. IMO, this makes the function names most accurate and concise. However, I'm a bit reluctant to suggest routing all the current `get_string` calls through my new function.

```diff
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1511,7 +1511,7 @@ class PrettyTable:
     def copy(self):
         return copy.deepcopy(self)
 
-    def get_output(self, out_format: str = "text", **kwargs) -> str:
+    def get_string(self, out_format: str = "text", **kwargs) -> str:
         """Return string representation of specified format of table in current state.
 
         Arguments:
@@ -1519,7 +1519,7 @@ class PrettyTable:
         kwargs - passed through to function that performs formatting
         """
         if out_format == "text":
-            return self.get_string(**kwargs)
+            return self.get_text_string(**kwargs)
         if out_format == "html":
             return self.get_html_string(**kwargs)
         if out_format == "json":
@@ -1692,7 +1692,7 @@ class PrettyTable:
     # PLAIN TEXT STRING METHODS  #
     ##############################
 
-    def get_string(self, **kwargs) -> str:
+    def get_text_string(self, **kwargs) -> str:
         """Return string representation of table in current state.
 
         Arguments:
```
All the unit tests pass (after I make adjustments to the tests that I introduced).

src/prettytable/prettytable.py
rickwporter: I also think the original is clearer despite being a little longer. If this was a "table" that was going to grow much, I would be much more likely to endorse the more elegant solution.

src/prettytable/prettytable.py
hugovk: The important thing would be to ensure `get_string()` remains fully backwards compatible.

Because the old `get_string()` only uses `**kwargs` and no positional arguments, we might be safe? We could make `out_format` a keyword only as well.

---

An alternative could be to avoid renaming or adding functions. Instead add `out_format` to the existing `get_string()`, and immediately check its value and potentially return one of the other (json, csv, etc) functions.

I'm less sure of this from a documentation point of view: `get_string()`'s docstring lists lots of args which don't apply to ther other functions.

rickwporter: I did not initially make the change to rename `get_string` to `get_text_string` because I was not 100% sure that none of the libraries called in those functions do NOT have an argument that shadows whatever name I choose. I figure the unit tests are a good way of testing, but not a guarantee. 

I guess the safest way to do this is introduce the new function that calls the existing function. I have renamed `get_output()` to `get_formatted_string()` to be more accurate (as suggested by @adilhussain540).

hugovk: > I did not initially make the change to rename `get_string` to `get_text_string` because I was not 100% sure that none of the libraries called in those functions do NOT have an argument that shadows whatever name I choose. I figure the unit tests are a good way of testing, but not a guarantee.

I wouldn't worry too much about that, the existing `get_*string` have dozens of arguments. And there's only a couple of libraries to check (csv and json?). If we wanted to be really careful we could call it `prettytable_format` or something.

But adding a new function is fine too.

adilhussain540 (approved):

hugovk:
Please could you document `get_formatted_string` somewhere in the README?

rickwporter:
> Please could you document `get_formatted_string` somewhere in the README?

Good call. I'm open to suggestions on the README. I was not quite sure where it best fit, or if my description length and content were appropriate.

README.md
```diff
@@ -203,6 +203,17 @@ This string is guaranteed to look exactly the same as what would be printed by d
 `print(x)`. You can now do all the usual things you can do with a string, like write
 your table to a file or insert it into a GUI.
 
+The table can be displayed in several different formats using `get_formatted_string`
+just by changing the `out_format=<text|html|json|csv|latex>`. This function passes
```
hugovk: Thanks! Let's remove a couple of words, re: https://justsimply.dev/

```suggestion
by changing the `out_format=<text|html|json|csv|latex>`. This function passes
```

README.md
```diff
@@ -203,6 +203,17 @@ This string is guaranteed to look exactly the same as what would be printed by d
...
+through arguments to the functions that render the table, so additional arguments
+can be given. The provides an easy means to let a user choose the output formatting.
```
```suggestion
can be given. This provides a way to let a user choose the output formatting.
```

hugovk (approved):
Thank you!

Issue #237, opened 2023-04-10
Dividers don't render correctly at cell intersections with other styles


con-f-use:
### What did you do?

Set a [row divider](https://github.com/jazzband/prettytable/pull/185) using an alternate style and print the resulting table.

### What did you expect to happen?

```terminal
$ python t.py
┌─────────┬─────────┬─────────┬─────────┐
│ Field 1 │ Field 2 │ Field 3 │ Field 4 │
├─────────┼─────────┼─────────┼─────────┤
│    11   │    12   │    13   │    14   │
│    21   │    22   │    23   │    24   │
│    31   │    32   │    33   │    34   │
├─────────┼─────────┼─────────┼─────────┤
│    41   │    42   │    43   │    44   │
└─────────┴─────────┴─────────┴─────────┘
```

### What actually happened?

```terminal
$ python t.py
┌─────────┬─────────┬─────────┬─────────┐
│ Field 1 │ Field 2 │ Field 3 │ Field 4 │
├─────────┼─────────┼─────────┼─────────┤
│    11   │    12   │    13   │    14   │
│    21   │    22   │    23   │    24   │
│    31   │    32   │    33   │    34   │
└─────────┴─────────┴─────────┴─────────┘  <---- should be three-way intersection
│    41   │    42   │    43   │    44   │
└─────────┴─────────┴─────────┴─────────┘
```

Note the transition into the row below the 2nd divider. The whole 2nd divider should look like the first one (after the "fields" row).

### What versions are you using?

* OS: NixOS 22.11 (Linux)
* Python:  3.10.10
* PrettyTable: 3.7.0

```python
# t.py
import prettytable

for style in [
    "DEFAULT", "MSWORD_FRIENDLY", "PLAIN_COLUMNS", "MARKDOWN",
    "ORGMODE", "DOUBLE_BORDER", "SINGLE_BORDER"
]:
    t = prettytable.PrettyTable()
    t.set_style(getattr(prettytable, style))
    t.add_row([11, 12, 13, 14])
    t.add_row([21, 22, 23, 24])
    t.add_row([31, 32, 33, 34])  # also happens with `divider=True`
    t.add_row([41, 42, 43, 44])
    t._dividers[2] = True  # gentle nudge that we need a better non-private API to set dividers after the fact
    print(t, style, "\n\n")     
```

hugovk:
To be explicit, here's a reproducer that doesn't use the internal/private `_dividers`:

```python
import prettytable

[...]

PR #263, opened 2023-10-17
Fixing error number #237

rajarshidcoder:
This bug occurred when we are trying to add a divider and changing the style of it such as "Double border" and it will produce unexpected results.


hugovk:
Please:

* Change the PR title to something descriptive. 
* Add tests that fail without the fix and passes with the fix.

hugovk:
@rajarshidcoder Hello! Reminder this PR is awaiting an update :)

hugovk:
Thanks anyway for the PR, I've updated this in https://github.com/prettytable/prettytable/pull/356.

PR #264, opened 2023-10-17
Add a feature to look like Excel

rajarshidcoder:
Adding this feature where user can choose this to be True so the final result will look like a Spreadsheet , or in short, will add a divider after every row.


hugovk:
Please can you give some example code and example output of this feature?

hugovk:
Closing due to lack of response.

Issue #251, opened 2023-07-08
Adjusting markdown table formatting to comply with GitHub flavored markdown specification


estyxx:
### Is your feature request related to a problem? Please describe.

When using prettytable to generate tables in the Markdown format, I've observed that the line separating the headers from the table body doesn't follow the standard [GitHub Flavored Markdown Spec](https://github.github.com/gfm/#tables-extension-). In the GitHub spec, there are spaces before and after the colons (`:`), like so:

```markdown
| :---- | :---------------------------------------- | -------: |
```

However, prettytable generates the line without these spaces:

```
|:-------|:-------------------------------------------|----------:|
```

### Describe the solution you'd like
I propose that prettytable be adjusted to adhere to the GitHub Flavored Markdown Spec for tables, which includes having spaces before and after the colons in the line separating headers from the table body.

### Describe alternatives you've considered
A possible alternative could be to reformat the generated tables manually, but this is not an efficient solution, especially when dealing with larger tables or multiple instances. It would be much more convenient and efficient for the formatting to be correct upon generation.

### Additional context
This formatting discrepancy can lead to inconsistencies when the tables are rendered in different Markdown parsers that strictly follow the GitHub Flavored Markdown Spec, which could affect the readability of the tables. Having prettytable adhere to this spec would eliminate such issues.

hugovk:
Sounds reasonable, would you like to put together a PR?

---

As an aside, interesting to see the GFM spec doesn't say that spaces should be part of the delimiter row:

> The [delimiter row](https://github.github.com/gfm/#delimiter-row) consists of cells whose only content are hyphens (-), and optionally, a leading or trailing colon (:), or both, to indicate left, right, or center alignment respectively.

And the only mention of spaces:

> Spaces between pipes and cell content are trimmed. 

But the examples clearly do include spaces.

I wonder if other Markdown variants are different? At least https://michelf.ca/projects/php-markdown/extra/#table also has spaces in the examples, and anyway, GFM is definitely a popular one.

firai:
> At least https://michelf.ca/projects/php-markdown/extra/#table also has spaces in the examples

By the way, the one example in Markdown Extra with a colon actually doesn't have a space after the colon, unlike the other examples without colons. However, both GFM and MultiMarkdown show spaces before and after colons.

PR #268, opened 2023-12-12
Use GitHub Flavored Markdown for table formatting

mumblingMac:
Fixed markdown table formatting to comply with GitHub Flavored Markdown Specification following the details provided in issue #251.

Updated the targeted unit tests as necessary.

Manually validated the generated table could be rendered using the Live Github-Flavored Markdown Editor.


hugovk:
Thank you!

PR #282, opened 2024-02-18
workflows/test.yml: bump coverage-action to v4

moto-timo:
https://github.com/codecov/codecov-action/releases/tag/v4.0.0

https://github.com/codecov/codecov-action/compare/v3.1.5...v4.0.1

Amongst other things, this also addresses the Nodejs 16 deprecation which was the reason for pinning to v3.1.5

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
           tox -e py
 
       - name: Upload coverage
-        uses: codecov/codecov-action@v3.1.5
+        uses: codecov/codecov-action@v4
         with:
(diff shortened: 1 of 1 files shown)


moto-timo:
I missed the comment in the dependabot v4 bump:
https://github.com/jazzband/prettytable/pull/275#issuecomment-1920554223

hugovk:
Thanks, yes, I think it's better to stick on v3 for a while, to see if things change.

moto-timo:
Agreed.

moto-timo:
What might be helpfull is if jazzband could have a organization wide codecov token... but I'm no codecov expert.

PR #286, opened 2024-02-27
Simplify empty dictionary/None check

himalayanZephyr:
Simplifying empty dictionary/None check

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -701,7 +701,7 @@ class PrettyTable:
 
     @min_width.setter
     def min_width(self, val) -> None:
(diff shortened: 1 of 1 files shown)


hugovk:
And change the other six occurrences?

We can then also flip from:

```python
if not val:
    ... # do when no val
else:
    ... # do when val
```

To:

```python
if val:
    ... # do when val
else:
    ... # do when no val
```

hugovk:
And can these be called when `val` is something that is _not_ `None` or `dict`, but still evaluates to false here? And if so, does it behave correctly?

hugovk:
@himalayanZephyr, please can you check the questions above?

hugovk:
Continued in https://github.com/prettytable/prettytable/pull/395.

PR #289, opened 2024-03-03
Fix max table width calculation

### LINKED ISSUE #288: max_table_width does not work as expected
### What did you do?
Set max_table_width on a table to `os.get_terminal_size().columns` and played around with terminal size.
 
The table should not exceed max_table_width (no line breaks in terminal)

max_width_should be max. 

### What actually happened?
Sometimes tables are wider than terminal size and i see line breaks

* OS: MacOs 
* Python: 3.12
* PrettyTable: 3.10.0

stuertz:
Try to fix max_table_width computation

fixes: #288

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1569,11 +1569,7 @@ class PrettyTable:
             if not options["fields"] or (
                 options["fields"] and fieldname in options["fields"]
             ):
-                table_width += (
-                    self._widths[index]
-                    + per_col_padding
-                    + (1 if options["vrules"] == ALL else 0)
-                )
+                table_width += self._widths[index] + per_col_padding + 1
         return table_width
 
     def _compute_widths(self, rows, options) -> None:
@@ -1605,12 +1601,7 @@ class PrettyTable:
             table_width = self._compute_table_width(options)
             if table_width > self._max_table_width:
                 # Shrink widths in proportion
-                if options["vrules"] == FRAME:
-                    vrulers = 2
-                elif options["vrules"] == ALL:
-                    vrulers = len(widths) - 1
-                else:
-                    vrulers = 0
+                vrulers = len(widths) - 1
                 scale = (
                     self._max_table_width - per_col_padding * len(widths) - vrulers
                 ) / (table_width - per_col_padding * len(widths) - vrulers)
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1605,8 +1605,7 @@ class PrettyTable:
                 scale = (
                     self._max_table_width - per_col_padding * len(widths) - vrulers
                 ) / (table_width - per_col_padding * len(widths) - vrulers)
-                widths = [max(1, int(w * scale)) for w in widths]
-                self._widths = widths
+                self._widths = [max(1, int(w * scale)) for w in widths]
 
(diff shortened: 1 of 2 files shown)


stuertz:
Hi @hugovk  is something missing here? Are you waiting for something to be added?

hugovk:
Hi! Thank you very much for the PR! I'm just waiting to find time to review it :) I'll definitely try to get to it this month.

src/prettytable/prettytable.py
```diff
@@ -1590,21 +1595,23 @@ def _compute_widths(self, rows, options) -> None:
...
             if table_width > self._max_table_width:
                 # Shrink widths in proportion
-                scale = 1.0 * self._max_table_width / table_width
-                widths = [int(w * scale) for w in widths]
+                vrulers = len(widths) - 1
+                scale = (
+                    self._max_table_width - per_col_padding * len(widths) - vrulers
+                ) / (table_width - per_col_padding * len(widths) - vrulers)
+                widths = [max(1, int(w * scale)) for w in widths]
                 self._widths = widths
```
hugovk: We can combine these:
```suggestion
                self._widths = [max(1, int(w * scale)) for w in widths]
```

src/prettytable/prettytable.py
```diff
@@ -1590,21 +1595,23 @@ def _compute_widths(self, rows, options) -> None:
...
 
+        per_col_padding = sum(self._get_padding_widths(options))
         # Are we exceeding max_table_width?
         if self._max_table_width:
             table_width = self._compute_table_width(options)
             if table_width > self._max_table_width:
                 # Shrink widths in proportion
-                scale = 1.0 * self._max_table_width / table_width
-                widths = [int(w * scale) for w in widths]
+                vrulers = len(widths) - 1
```
hugovk: Can you think of another name for `vrulers`? It's not clear from the name what it's for.

src/prettytable/prettytable.py
```diff
@@ -1590,21 +1595,23 @@ def _compute_widths(self, rows, options) -> None:
...
+                scale = (
+                    self._max_table_width - per_col_padding * len(widths) - vrulers
+                ) / (table_width - per_col_padding * len(widths) - vrulers)
```
hugovk: Both numerator and denominator use `per_col_padding * len(widths) - vrulers`.

Shall we move this out to a variable pre-calculating `per_col_padding * len(widths) + vrulers`?

Won't really affect performance, but will make the `scale` line easier to read.

We could also pre-calculate as `per_col_padding * len(widths) + len(widths) - 1` and remove `vrulers`.

hugovk (commented):
Thank you for this! This is mainly style/readability suggestions.

src/prettytable/prettytable.py
stuertz: It's the number of vertical chars or how do you call `|`. What about `no_of_vchars`

hugovk: Sure, `no_of_vchars` or `num_vchars` or `n_vchars` or something?

src/prettytable/prettytable.py
stuertz: yep. I'll chose `markup_chars` as varname for the param

hugovk:
Thank you very much!

PR #299, opened 2024-04-19
Change how size of headers are calculated to fix ColorTable with title

### LINKED ISSUE #290: Colortable with title leads to broken layout
### What did you do?
printed a colortable with a title

normal layout as with a prettytable, but colored

### What actually happened?
title line explodes in width and one escape sequence was also printed

* OS:  macos
* Python:  3.12
* PrettyTable:  3.10.0

```python
from prettytable import colortable

table = colortable.ColorTable(
    ("A", "B", "C", "D", "E", "f"),
    align="l",
    theme=colortable.Themes.OCEAN,
)
table.title = "Efforts"
table.add_row([1, 2, 3, 4, 5, 6])
print(table)
```

Creates (colored):
```console
+-----------------------+
|                                                                                                                                                                                       Efforts                                                                                                                                                                                       |
+[36m+---+---+---+---+---+---++
| A | B | C | D | E | f |
+---+---+---+---+---+---+
| 1 | 2 | 3 | 4 | 5 | 6 |
+---+---+---+---+---+---+
```

av-guy:
Fixes #290

The issue that the `ColorTable` is experiencing deals with its incorporation of ANSI escape character sequences. The `ColorTable` class works by incorporating ANSI escape character sequences both before and after the character it represents. That creates a discrepancy in how `PrettyTable` and `ColorTable` portray the same character.

I altered how `_stringify_title` computes its width. Instead of relying on `self._hrule`, it now uses the padding widths and the widths of the columns themselves. From my testing, this seems to work well. I also modified the way `_stringify_header` slices the `bits` value. The old code used hard values of `1` and `-1`, which doesn't account for the length of a single character when someone uses `ColorTable`.

If you see anything that stands out, or you have input on a different solution, please let me know. I am open to the discussion. Also, I didn't see any tests related to the table rendering for `ColorTable`, so I didn't add any. All of the original tests are passing for both `ColorTable` and `PrettyTable` with these recent changes.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1888,7 +1888,7 @@ class PrettyTable:
         bits.append(endpoint)
         title = " " * lpad + title + " " * rpad
         lpad, rpad = self._get_padding_widths(options)
-        sum_widths = sum(n + lpad + rpad + 1 for n in self._widths)
+        sum_widths = sum([n + lpad + rpad + 1 for n in self._widths])
 
         bits.append(self._justify(title, sum_widths - 1, "c"))
         bits.append(endpoint)
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1888,7 +1888,7 @@ class PrettyTable:
         bits.append(endpoint)
         title = " " * lpad + title + " " * rpad
         lpad, rpad = self._get_padding_widths(options)
-        sum_widths = sum((n + lpad + rpad + 1 for n in self._widths))
+        sum_widths = sum(n + lpad + rpad + 1 for n in self._widths)
 
         bits.append(self._justify(title, sum_widths - 1, "c"))
         bits.append(endpoint)
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1888,9 +1888,9 @@ class PrettyTable:
         bits.append(endpoint)
         title = " " * lpad + title + " " * rpad
         lpad, rpad = self._get_padding_widths(options)
-        total_widths = [n + lpad + rpad + 1 for n in self._widths]
+        sum_widths = sum((n + lpad + rpad + 1 for n in self._widths))
 
-        bits.append(self._justify(title, sum(total_widths) - 1, "c"))
+        bits.append(self._justify(title, sum_widths - 1, "c"))
         bits.append(endpoint)
         lines.append("".join(bits))
         return "\n".join(lines)
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1887,7 +1887,10 @@ class PrettyTable:
         )
         bits.append(endpoint)
         title = " " * lpad + title + " " * rpad
-        bits.append(self._justify(title, len(self._hrule) - 2, "c"))
+        lpad, rpad = self._get_padding_widths(options)
+        total_widths = [n + lpad + rpad + 1 for n in self._widths]
+
+        bits.append(self._justify(title, sum(total_widths) - 1, "c"))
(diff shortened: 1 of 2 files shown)


src/prettytable/prettytable.py
```diff
@@ -1887,7 +1887,10 @@ def _stringify_title(self, title, options):
         )
         bits.append(endpoint)
         title = " " * lpad + title + " " * rpad
-        bits.append(self._justify(title, len(self._hrule) - 2, "c"))
+        lpad, rpad = self._get_padding_widths(options)
+        total_widths = [n + lpad + rpad + 1 for n in self._widths]
+
+        bits.append(self._justify(title, sum(total_widths) - 1, "c"))
```
hugovk: Creating and summing a tuple is nearly 2x as fast as a list:

```suggestion
        sum_widths = sum((n + lpad + rpad + 1 for n in self._widths))

        bits.append(self._justify(title, sum_widths - 1, "c"))
```

hugovk (changes requested):
Thanks for the PR!

Please could you:

* Update the title to be more descriptive? It will be used in the release notes. See https://github.com/jazzband/prettytable/releases to get an idea.

* Add "Fixes #xxx" with the issue number to the PR description, so the issue will be closed when this is merged.

* Add tests. It's good none of the existing tests fail. It will be even better to add tests which fail with current `main` and pass with this PR, so we can avoid regressions in the future.

av-guy:
> Thanks for the PR!
> 
> Please could you:
> 
> * Update the title to be more descriptive? It will be used in the release notes. See https://github.com/jazzband/prettytable/releases to get an idea.
> * Add "Fixes #xxx" with the issue number to the PR description, so the issue will be closed when this is merged.
> * Add tests. It's good none of the existing tests fail. It will be even better to add tests which fail with current `main` and pass with this PR, so we can avoid regressions in the future.

I will try to get the tests done today. I did change the PR description. Does the new description satisfy the first two requirements in your list?

hugovk:
Looks good, except please move "Fixes #xxx" from the title to PR the description.

av-guy:
> Looks good, except please move "Fixes #xxx" from the title to PR the description.

I added Fixes #290 to my original comment. I don't contribute very often, so if I did this wrong, please let me know.

av-guy:
> Looks good, except please move "Fixes #xxx" from the title to PR the description.

I added a basic test for the color table. Please see commit df8f2e6 and provide feedback as necessary.

src/prettytable/prettytable.py
```diff
@@ -1887,7 +1887,10 @@ def _stringify_title(self, title, options):
         )
         bits.append(endpoint)
         title = " " * lpad + title + " " * rpad
-        bits.append(self._justify(title, len(self._hrule) - 2, "c"))
+        lpad, rpad = self._get_padding_widths(options)
+        sum_widths = sum(n + lpad + rpad + 1 for n in self._widths)
```
hugovk: Actually, my last suggestion was wrong, and summing a list comprehension is faster than a generator:

```console
❯ python -m timeit -s "a = (1,2,3,4,4,6,7,8,9)" "sum(tuple(n for n in a))"
1000000 loops, best of 5: 363 nsec per loop

❯ python -m timeit -s "a = (1,2,3,4,4,6,7,8,9)" "sum(n for n in a)"
1000000 loops, best of 5: 270 nsec per loop

❯ python -m timeit -s "a = (1,2,3,4,4,6,7,8,9)" "sum([n for n in a])"
2000000 loops, best of 5: 168 nsec per loop
```


```suggestion
        sum_widths = sum([n + lpad + rpad + 1 for n in self._widths])
```

hugovk (approved):
Thank you!

PR #114, opened 2021-05-27
Add Escaping Options to HTML Tables

geekiegirl2016:
Fixes #40. 

Add Escaping Options to HTML Tables

This will allow tables to contain active links and other html code allowing prettytables to more easily be used for dynamic site building.

Example includes allowing a header to be a link.

`x.get_html_string(header_escape=False, data_escape=False)`

```html
<table>
    <thead>
        <tr>
            <th><a href="https://google.com">Field 1</a></th>
            <th>Field 2</th>
        </tr>
    </thead>
        <tr>
            <td>value 1</td>
            <td>value 2</td>
        </tr>
        <tr>
            <td><a href="https://google.com">value 3</a></td>
            <td>value 4</td>
        </tr>
    </tbody>
</table>
```

instead of
`x.get_html_string()`

```html
<table>
    <thead>
        <tr>
            <th>&lt;a href=&quot;https://google.com&quot;&gt;Field 1&lt;/a&gt;</th>
            <th>Field 2</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>value 1</td>
            <td>value 2</td>
        </tr>
        <tr>
            <td>&lt;a href=&quot;https://google.com&quot;&gt;value 3&lt;/a&gt;</td>
            <td>value 4</td>
        </tr>
    </tbody>
</table>
```

PR #305, opened 2024-07-31
Add options to disable escaping HTML tables

### LINKED ISSUE #40: Make escaping data in an HTML table optional
For certain rows, I require the data to be augmented with HTML, and that unfortunately gets escaped here: https://github.com/jazzband/prettytable/blob/master/prettytable.py#L1622

Would you consider adding an option to get_html_string to disable data escaping?

Happy to contribute a PR

MrMarble:
Fixes #40

Since #114 was closed, I decided to replicate the changes as this is a very needed feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -675,6 +675,16 @@ will print:
 </table>
 ```
 
+#### Setting HTML Escaping
+
+By default, PrettyTable will escape the data contained in the header and data fields
+when sending output to html. This can be disabled by setting the 'escape_header' and
+'escape_data' to false. For example:
+
+```python
+print(x.get_html_string(escape_header=False, escape_data=False))
+```
+
 ### Miscellaneous things
 
 #### Copying a table
--- a/README.md
+++ b/README.md
@@ -675,7 +675,7 @@ will print:
 </table>
 ```
 
-#### Setting HTML Escaping
+#### Setting HTML escaping
 
 By default, PrettyTable will escape the data contained in the header and data fields
 when sending output to HTML. This can be disabled by setting the `escape_header` and
--- a/README.md
+++ b/README.md
@@ -678,11 +678,11 @@ will print:
 #### Setting HTML Escaping
 
 By default, PrettyTable will escape the data contained in the header and data fields
-when sending output to html. This can be disabled by setting the 'escape_header' and
-'escape_data' to false. For example:
+when sending output to HTML. This can be disabled by setting the `escape_header` and
+`escape_data` to false. For example:
 
 ```python
-print(x.get_html_string(escape_header=False, escape_data=False))
+print(table.get_html_string(escape_header=False, escape_data=False))
 ```
 
 ### Miscellaneous things
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2232,15 +2232,12 @@ class PrettyTable:
                 if options["fields"] and field not in options["fields"]:
                     continue
                 if options["escape_header"]:
-                    lines.append(
-                        "            <th>{}</th>".format(
-                            escape(field).replace("\n", linebreak)
-                        )
-                    )
-                else:
-                    lines.append(
-                        "            <th>{}</th>".format(field.replace("\n", linebreak))
-                    )
+                    field = escape(field)
+
+                lines.append(
+                    "            <th>{}</th>".format(field.replace("\n", linebreak))
+                )
+
             lines.append("        </tr>")
             lines.append("    </thead>")
 
@@ -2254,15 +2251,11 @@ class PrettyTable:
                 if options["fields"] and field not in options["fields"]:
                     continue
                 if options["escape_data"]:
-                    lines.append(
-                        "            <td>{}</td>".format(
-                            escape(datum).replace("\n", linebreak)
-                        )
-                    )
-                else:
-                    lines.append(
-                        "            <td>{}</td>".format(datum.replace("\n", linebreak))
-                    )
+                    datum = escape(datum)
+
+                lines.append(
+                    "            <td>{}</td>".format(datum.replace("\n", linebreak))
+                )
             lines.append("        </tr>")
         lines.append("    </tbody>")
         lines.append("</table>")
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -675,6 +675,16 @@ will print:
 </table>
 ```
 
+#### Setting HTML Escaping
+
+By default, PrettyTable will escape the data contained in the header and data fields
+when sending output to html. This can be disabled by setting the 'escape_header' and
+'escape_data' to false. For example:
```
```suggestion
By default, PrettyTable will escape the data contained in the header and data fields
when sending output to HTML. This can be disabled by setting the `escape_header` and
`escape_data` to false. For example:
```

README.md
```diff
@@ -675,6 +675,16 @@ will print:
...
+
+```python
+print(x.get_html_string(escape_header=False, escape_data=False))
```
```suggestion
print(table.get_html_string(escape_header=False, escape_data=False))
```

MrMarble:
Is this planned for a release soon?

src/prettytable/prettytable.py
```diff
@@ -2197,11 +2231,16 @@ def _get_simple_html_string(self, options):
...
+                if options["escape_header"]:
+                    lines.append(
+                        "            <th>{}</th>".format(
+                            escape(field).replace("\n", linebreak)
+                        )
+                    )
+                else:
+                    lines.append(
+                        "            <th>{}</th>".format(field.replace("\n", linebreak))
                     )
```
hugovk: Can we simplify this?
```suggestion

                if options["escape_header"]:
                    field = escape(field)

                lines.append(
                    "            <th>{}</th>".format(field.replace("\n", linebreak))
                )
```


But with this version, or the current PR version, the tests pass even if we replace `escape(field)` with `field`. Please can you check this is tested properly?

hugovk:
Let's review and merge it first :)

After that we can release pretty much whenever. Review underway.

src/prettytable/prettytable.py
```diff
@@ -2214,11 +2253,16 @@ def _get_simple_html_string(self, options):
...
+                if options["escape_data"]:
+                    lines.append(
+                        "            <td>{}</td>".format(
+                            escape(datum).replace("\n", linebreak)
+                        )
+                    )
+                else:
+                    lines.append(
+                        "            <td>{}</td>".format(datum.replace("\n", linebreak))
                     )
```
```suggestion

                if options["escape_data"]:
                    datum = escape(datum)

                lines.append(
                    "            <td>{}</td>".format(datum.replace("\n", linebreak))
                )
```

Same thing about the tests.

src/prettytable/prettytable.py
```diff
@@ -2273,10 +2317,16 @@ def _get_formatted_html_string(self, options):
...
+                if options["escape_header"]:
+                    lines.append(
+                        '            <th style="padding-left: %dem; padding-right: %dem; text-align: center">%s</th>'  # noqa: E501
+                        % (lpad, rpad, escape(field).replace("\n", linebreak))
+                    )
+                else:
+                    lines.append(
+                        '            <th style="padding-left: %dem; padding-right: %dem; text-align: center">%s</th>'  # noqa: E501
+                        % (lpad, rpad, field.replace("\n", linebreak))
+                    )
```
```suggestion

                if options["escape_header"]:
                    field = escape(field)

                lines.append(
                    '            <th style="padding-left: %dem; padding-right: %dem; text-align: center">%s</th>'  # noqa: E501
                    % (lpad, rpad, field.replace("\n", linebreak))
                )
```
And same about tests here.

src/prettytable/prettytable.py
```diff
@@ -2300,16 +2350,28 @@ def _get_formatted_html_string(self, options):
...
+                    lines.append(
+                        '            <td style="padding-left: %dem; padding-right: %dem; text-align: %s; vertical-align: %s">%s</td>'  # noqa: E501
+                        % (
+                            lpad,
+                            rpad,
+                            align,
+                            valign,
+                            datum.replace("\n", linebreak),
+                        )
                     )
```
```suggestion

                if options["escape_data"]:
                    datum = escape(datum)

                lines.append(
                    '            <td style="padding-left: %dem; padding-right: %dem; text-align: %s; vertical-align: %s">%s</td>'  # noqa: E501
                    % (
                        lpad,
                        rpad,
                        align,
                        valign,
                        datum.replace("\n", linebreak),
                    )
                )
```
And here.

src/prettytable/prettytable.py
MrMarble: I added two new tests that checks the current escaped output

src/prettytable/prettytable.py
```diff
@@ -2214,10 +2250,11 @@ def _get_simple_html_string(self, options):
             for field, datum in zip(self._field_names, row):
                 if options["fields"] and field not in options["fields"]:
                     continue
+                if options["escape_data"]:
+                    datum = escape(datum)
```
hugovk, line 2254: Tests still pass without this.

src/prettytable/prettytable.py
```diff
@@ -2273,9 +2310,12 @@ def _get_formatted_html_string(self, options):
             for field in self._field_names:
                 if options["fields"] and field not in options["fields"]:
                     continue
+                if options["escape_header"]:
+                    field = escape(field)
```
hugovk, line 2314: And without this.

src/prettytable/prettytable.py
MrMarble: What do you mean?
If I remove that, the tests fails for me
image
image

hugovk: Sorry, I meant line 2344.

src/prettytable/prettytable.py
MrMarble: This does pass, because the method `_get_formatted_html_string` is not tested.
Should I add another 4 tests with the `result = t.get_html_string(escape_header=False,format=True)`?

hugovk: Yes please. It would be good to make sure some future change doesn't accidentally break these changes.

I've not looked closely at the tests yet, I think we could just use something like a single header row and single body row, to make the tests more compact. We can pass the number of rows to `t = helper_table(rows=1)`. And maybe `rows=0` where you append a row.

hugovk: Thanks for the update! It's looking good, I'll try and give it a final review this weekend and we can hopefully merge it soon!

README.md
```diff
@@ -675,6 +675,16 @@ will print:
 </table>
 ```
 
+#### Setting HTML Escaping
```
```suggestion
#### Setting HTML escaping
```

hugovk (approved):
Thank you!

PR #307, opened 2024-08-04
Add field selection to CSV and JSON outputs

### LINKED ISSUE #253: Support displaying selected fields for CSV format
When calling `get_string` or `get_html_string` or `get_latex_string` and passing `fields = ["field1", "field2"]` as the parameter, the output strings in either format contain the given fields in `fields` only. However, when calling `get_csv_string` or `get_json_string`, that does not work. It seems that these two functions do not take this setting into consideration.

demestav:
Closes #253 

When the `fields` option is passed to `get_csv_string` and `get_json_string`, the output will include only the specified fields.


hugovk:
Thanks!

PR #313, opened 2024-09-27
doc: mypy-compliant type for column alignement

dirtyhillbilly:
Using property for `align` confuses mypy type checker.

This MR updates the documentation to make clear how to properly use `table.align[fieldname]`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,10 +299,10 @@ You can also change the alignment of individual columns based on the correspondi
 name by treating the `align` attribute as if it were a dictionary.
 
 ```python
-table.align["City name"] = "l"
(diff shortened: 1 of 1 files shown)


hugovk:
How does this confuse mypy?

dirtyhillbilly:
The following code was hiting a `Unsupported target for indexed assignment`
```
for name, type_ in sample.items():
  table.align[name] = "l" if issubclass(str, type_) else "r"
```

But for some reason , now that it is fixed, I can't reproduce the error (maybe a mypy update in poetry venv or something...).

You can close this MR, sorry for the noise.

PR #316, opened 2024-10-12
Add attribute type hints

cdce8p:
Small step to improve the type hints. Add annotations for basic attributes. They might not be 100% correct but it's always possible to refine them later.

Also removed the `types-setuptools` dependency since `setuptools` is no longer used.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -66,6 +66,36 @@ def _get_size(text):
 
 
 class PrettyTable:
+    _xhtml: bool
+    _min_table_width: int | None
+    _max_table_width: int | None
+    _title: str | None
+    _start: int
+    _end: int | None
+    _reversesort: bool
+    _header: bool
+    _border: bool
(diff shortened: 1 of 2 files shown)


src/prettytable/prettytable.py
```diff
@@ -1030,13 +1060,13 @@ def vertical_char(self):
         return self._vertical_char
 
     @vertical_char.setter
-    def vertical_char(self, val) -> None:
+    def vertical_char(self, val: str) -> None:
         val = str(val)
         self._validate_option("vertical_char", val)
```
cdce8p, line 1065: A lot of these attribute setters could work with a wider range of values. To be useful for type checkers, I think it makes sense to use the closest matching type here though.

E.g. even if an `int` would work with `str(1)`, it makes sense if type checkers are able to warn if it isn't a string which is passed to the attribute setter.

hugovk (approved):
Thank you!

PR #317, opened 2024-10-13
Adjust format validation

cdce8p:
The docstring for `format` says it should be either `True` or `False`. Change the validator from `_validate_nonnegative_int` to `_validate_true_or_false` to better reflect that.

It succeeded previously as `_validate_nonnegative_int` uses `int(value) >= 0`.


hugovk:
Thanks again!

PR #319, opened 2024-10-13
Improve function type annotations

cdce8p:
Add some strait forward type annotations to functions and methods.

> [!Note]
> Explicitly leave out the more complicated stuff to make it easier to review.
>
> Unfortunately though, quite a few of the remaining untyped decorators require different type hints for the getter and setter. I could add the annotations, but this is currently unsupported by mypy. https://github.com/python/mypy/issues/3004

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2698,7 +2698,7 @@ def from_html(html_code: str, **kwargs) -> list[PrettyTable]:
 
 def from_html_one(html_code: str, **kwargs) -> PrettyTable:
     """
-    Generates a PrettyTables from a string of HTML code which contains only a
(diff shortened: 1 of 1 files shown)


src/prettytable/prettytable.py
```diff
@@ -2693,7 +2696,7 @@ def from_html(html_code, **kwargs):
     return parser.tables
 
 
-def from_html_one(html_code, **kwargs):
+def from_html_one(html_code: str, **kwargs) -> PrettyTable:
     """
     Generates a PrettyTables from a string of HTML code which contains only a
```
hugovk: Just spotted this little typo, let's slip a fix in:

```suggestion
    Generates a PrettyTable from a string of HTML code which contains only a
```

hugovk (approved):
Thanks!

PR #325, opened 2024-10-15
Add type aliases for `Align`, `VAlign`, `HeaderStyle`

cdce8p:


src/prettytable/prettytable.py
```diff
@@ -513,9 +519,9 @@ def _validate_align(self, val):
 
     def _validate_valign(self, val):
         try:
-            assert val in ["t", "m", "b", None]
+            assert val in ["t", "m", "b"]
         except AssertionError:
-            msg = f"Alignment {val} is invalid, use t, m, b or None"
+            msg = f"Alignment {val} is invalid, use t, m, b"
             raise ValueError(msg)
```
cdce8p, line 525: It's still possible to set `valign` to `None`. That's just handled by the property handler. The validation method will never get `None`.
https://github.com/jazzband/prettytable/blob/c7c89cf336755b3ed6662e4ad322bf980d9a06a6/src/prettytable/prettytable.py#L719-L729

--
That's similar to `_validate_align`.
https://github.com/jazzband/prettytable/blob/c7c89cf336755b3ed6662e4ad322bf980d9a06a6/src/prettytable/prettytable.py#L507-L512

PR #326, opened 2024-10-15
Add `IntEnum` for `Table` styles

cdce8p:


src/prettytable/prettytable.py
```diff
@@ -51,15 +52,29 @@
...
+    MSWORD_FRIENDLY = 11
+    PLAIN_COLUMNS = 12
+    MARKDOWN = 13
+    ORGMODE = 14
+    DOUBLE_BORDER = 15
+    SINGLE_BORDER = 16
+    RANDOM = 20
+
+
+# keep for backwards compatibility
```
hugovk, line 67: What do you think about deprecating these?

cdce8p: Fine by me. Wasn't sure what your police for that was.

Would you be ok if I do that in a followup? I'm currently working on a similar PR for `HRuleStyle` and `VRuleStyle`. Merging this now would prevent merge conflicts when I open that one.

hugovk: Sure, it can wait, let's merge this now.

PR #327, opened 2024-10-15
Add `IntEnum` for `HRuleStyle` and `VRuleStyle`

cdce8p:
Similar to #326, this time for `HRuleStyle` and `VRuleStyle`.

It's a bit larger than all the previous ones but it didn't really make sense to split it up into multiple PRs. Instead each commit can be best reviewed one after another.


src/prettytable/prettytable.py
```diff
@@ -1399,7 +1419,7 @@ def _set_orgmode_style(self) -> None:
     def _set_markdown_style(self) -> None:
         self.header = True
         self.border = True
-        self._hrules = None
+        self._hrules = HRuleStyle.HEADER
```
cdce8p, line 1415: `None` is not a "valid" value. The closest would be `NONE`, however that removes the header completely. Using `HRuleStyle.HEADER` doesn't require any test changes and is fully compatible with the previous `None`. https://github.com/jazzband/prettytable/pull/327/commits/06a644261ca7e0e7e2559e1cfe034d92dc9c4150

https://github.com/jazzband/prettytable/blob/d80f5234a7dc27c72c1c080b6afb76ab131cc89a/tests/test_prettytable.py#L1622-L1631

src/prettytable/prettytable.py
```diff
@@ -1476,8 +1496,8 @@ def _set_random_style(self) -> None:
 
         self.header = random.choice((True, False))
         self.border = random.choice((True, False))
-        self._hrules = random.choice((ALL, FRAME, HEADER, NONE))
-        self._vrules = random.choice((ALL, FRAME, NONE))
+        self._hrules = random.choice(list(HRuleStyle._value2member_map_))
+        self._vrules = random.choice(list(VRuleStyle._value2member_map_))
```
cdce8p: The order changes a bit and therefore also the output since the seed is kept the same. https://github.com/jazzband/prettytable/pull/327/commits/0c23205e62f5d73424e88eaca3a9a899af422a82

Technically a small behavior change, however as it's for `TableStyle.RANDOM` I think it's fine. The alternative would be to replace each constant with the corresponding enum value which would preserve the order.

cdce8p (commented):
Didn't intent to force-push that often. Working with `random` and missing the compatibility to old Python versions. _Been a while since I last used 3.11 😅_

Anyway, this is now ready for review.

src/prettytable/prettytable.py
hugovk: I think order change is okay too, although `_value2member_map_` isn't very clear. Can we keep the individual enums instead?

src/prettytable/prettytable.py
```diff
@@ -596,16 +610,22 @@ def _validate_function(self, name, val):
...
+            msg = f"Invalid value for {name}. Must be HRuleStyle."
             raise ValueError(msg)
 
     def _validate_vrules(self, name, val):
         try:
-            assert val in (ALL, FRAME, NONE)
+            if sys.version_info >= (3, 12):
+                assert val in VRuleStyle
+            else:
+                assert val in VRuleStyle._value2member_map_
```
hugovk: Also this one (and the next): `_value2member_map_` is undocumented, so is it "private" and open to change/removal with no notice?

cdce8p: I don't think it get's removed for Python `<=3.11`. If you like, I can change that as well though.

cdce8p: I might have found an even better solution with `val in list(VRuleStyle)`. That also works for `random.choice`. See https://github.com/jazzband/prettytable/pull/327/commits/084beaefa2c12aa8704c6a66079827f5a8f89574

hugovk: Much nicer!

PR #328, opened 2024-10-15
Small improvements

cdce8p:


src/prettytable/prettytable.py
```diff
@@ -239,7 +242,6 @@ def __init__(self, field_names: Sequence[str] | None = None, **kwargs) -> None:
             "horizontal_align_char",
             "junction_char",
             "header_style",
-            "valign",
```
cdce8p, line 242: `valign` is present twice in `self._options`.
https://github.com/jazzband/prettytable/blob/d80f5234a7dc27c72c1c080b6afb76ab131cc89a/src/prettytable/prettytable.py#L242-L255

src/prettytable/prettytable.py
```diff
@@ -96,6 +96,8 @@ class PrettyTable:
     _xhtml: bool
     _align: dict[str, AlignType]
     _valign: dict[str, VAlignType]
+    _max_width: dict[str, int]
+    _min_width: dict[str, int]
```
```suggestion
    _min_width: dict[str, int]
    _max_width: dict[str, int]
```
Same order as the following pair.

PR #329, opened 2024-10-16
Emit `DeprecationWarning`s for `hrule` and `tableStyle` constants

cdce8p:
Followup from https://github.com/jazzband/prettytable/pull/326#discussion_r1801633213


src/prettytable/__init__.py
```diff
@@ -27,18 +30,6 @@
...
-    "DOUBLE_BORDER",
-    "SINGLE_BORDER",
-    "FRAME",
-    "HEADER",
-    "MARKDOWN",
-    "MSWORD_FRIENDLY",
-    "NONE",
-    "ORGMODE",
-    "PLAIN_COLUMNS",
-    "RANDOM",
```
cdce8p: I removed the constants from `__all__` so type checks would emit warnings when there are used. The import is still possible though as to not break existing code.

**Edit**: Just realized that this would break `from prettytable import *`. I'll add them back then.

tests/test_prettytable.py
```diff
@@ -2606,3 +2597,51 @@ def test_invalid(self) -> None:
...
+    )
+    @pytest.mark.parametrize(
+        ["name"],
+        [
+            ("FRAME",),
+            ("ALL",),
+            ("NONE",),
+            ("HEADER",),
+        ],
+    )
```
```suggestion
    @pytest.mark.parametrize(
        "module_name",
        [
            "prettytable",
            "prettytable.prettytable",
        ],
    )
    @pytest.mark.parametrize(
        "name",
        [
            "FRAME",
            "ALL",
            "NONE",
            "HEADER",
        ],
    )
```

tests/test_prettytable.py
```diff
@@ -2606,3 +2597,51 @@ def test_invalid(self) -> None:
...
+            ("MSWORD_FRIENDLY",),
+            ("PLAIN_COLUMNS",),
+            ("MARKDOWN",),
+            ("ORGMODE",),
+            ("DOUBLE_BORDER",),
+            ("SINGLE_BORDER",),
+            ("RANDOM",),
+        ],
+    )
+    def test_tablestyle_constant_deprecations(
```
```suggestion
    @pytest.mark.parametrize(
        "module_name",
        [
            "prettytable",
            "prettytable.prettytable",
        ],
    )
    @pytest.mark.parametrize(
        "name",
        [
            "DEFAULT",
            "MSWORD_FRIENDLY",
            "PLAIN_COLUMNS",
            "MARKDOWN",
            "ORGMODE",
            "DOUBLE_BORDER",
            "SINGLE_BORDER",
            "RANDOM",
        ],
    )
    def test_table_style_constant_deprecations(
```

src/prettytable/prettytable.py
```diff
@@ -2791,3 +2792,22 @@ def from_html_one(html_code: str, **kwargs) -> PrettyTable:
...
+        msg = f"module '{__name__}' has no attribute '{name}"
+        raise AttributeError(msg)
+    if name in {"FRAME", "ALL", "NONE", "HEADER"}:
+        msg = (
+            f"the '{name}' constant is deprecated, "
+            "use the 'HRuleStyle' and 'VRuleStyle' enums instead"
+        )
+    else:
+        msg = f"the '{name}' constant is deprecated, use the 'TableStyle' enum instead"
+    warnings.warn(msg, DeprecationWarning, stacklevel=2)
```
```suggestion
    warnings.warn(msg, DeprecationWarning, stacklevel=3)
```

hugovk:
Thanks!

cdce8p:
Just tested it one more time. When I import a deprecated symbol from just `prettytable`, the warning is emitted twice. Do you have an idea how to resolve that?
```py
# test.py
from prettytable import ALL
```
```bash
$ python -Wall test.py 
/.../test.py:1: DeprecationWarning: the 'ALL' constant is deprecated, use the 'HRuleStyle' and 'VRuleStyle' enums instead
  from prettytable import ALL
/.../test.py:1: DeprecationWarning: the 'ALL' constant is deprecated, use the 'HRuleStyle' and 'VRuleStyle' enums instead
  from prettytable import ALL
```

cdce8p:
Ah too late. Anyhow, if you have an idea 🔝, I can open another PR for that.

hugovk:
Hmm, will have a look later on!

hugovk:
It happens that way because `__getattr__` is called twice for:

```python
from prettytable import ALL
```

But it's only called once for:

```python
import prettytable
prettytable.ALL
```

See https://stackoverflow.com/questions/60774515/python-getattr-executed-multiple-times

I'm not sure if it's worth trying to work around?

cdce8p:
> I'm not sure if it's worth trying to work around?

The fix was actually quite simple. Should have done from the beginning 😅
- #337

PR #332, opened 2024-10-17
Fix drawing headerless colored tables with title

kuba2k2:
The pull request #299 (issue #290) fixed drawing `ColorTable` with a title set. However, it didn't fix drawing such a table with `header` set to False.

This PR fixes that; there was some incorrect line slicing that led to ANSI codes being cut off partially.

Here's how it looks without this PR:
```
+-----------------------+
|        Efforts        |
+[36m+---+---+---+---+---+---+ 1 | 2 | 3 | 4 | 5 | 6 |
+---+---+---+---+---+---+
```
and with this PR - the correct result:
```
+-----------------------+
|        Efforts        |
+---+---+---+---+---+---+
| 1 | 2 | 3 | 4 | 5 | 6 |
+---+---+---+---+---+---+
```
Produced by this code:
```python
from prettytable import colortable

table = colortable.ColorTable(
    ("A", "B", "C", "D", "E", "f"),
    align="l",
    theme=colortable.Themes.OCEAN,
)
table.title = "Efforts"
table.header = False
table.add_row([1, 2, 3, 4, 5, 6])
print(table)
```

I have also added parametrized tests to the `test_colortable.py`, so that it will test all possible cases of header/title combinations.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1923,8 +1923,12 @@ class PrettyTable:
         ):
             lines.append(self._stringify_hrule(options, where="top_"))
             if title and options["vrules"] in (VRuleStyle.ALL, VRuleStyle.FRAME):
+                left_j_len = len(self.left_junction_char)
+                right_j_len = len(self.right_junction_char)
                 lines[-1] = (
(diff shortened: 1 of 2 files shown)


hugovk:
Thank you!

PR #336, opened 2024-10-26
Release to PyPI using Trusted Publishing

hugovk:
PyPI has introduced "Trusted Publishers", a method to release files from CI using OIDC and generated, short-lived tokens, rather than long-lived tokens on the developer's own machine. This is both safer, and makes releasing easier and more convenient.

https://docs.pypi.org/trusted-publishers/

This PR adds a workflow to deploy to PyPI for new GitHub releases.

It also deploys to Test PyPI on merges to `main`, to make sure the release machinery is well oiled.

I've set up the PyPIs with this:

image

* https://test.pypi.org/manage/project/prettytable/settings/publishing/
* https://pypi.org/manage/project/prettytable/settings/publishing/

---

PEP 740 ("Index support for digital attestations") introduces signatures which links the PyPI package to the GitHub repo, and helps users verify the source and authenticity of packages. This is only available with Trusted Publishing.

PyPI is still implementing support, but we can already start using it, which should also help them test out.

* https://peps.python.org/pep-0740/
* https://github.com/pypa/gh-action-pypi-publish#generating-and-uploading-attestations
* https://github.com/pypi/warehouse/issues/15871

All we need to do to enable this is add:
```yml
        with:
          attestations: true
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,79 @@
+name: Build package
+
+on:
+  push:
+  pull_request:
+  release:
+    types:
+      - published
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+env:
+  FORCE_COLOR: 1
+
+jobs:
+  # Always build & lint package.
+  build-package:
+    name: Build & verify package
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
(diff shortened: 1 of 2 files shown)


hugovk:
Let's merge and check if `main` -> Test PyPI works.

hugovk:
🎯 https://github.com/prettytable/prettytable/actions/runs/11533000950/job/32105417894 -> https://test.pypi.org/project/prettytable/3.11.1.dev29/

PR #338, opened 2024-10-29
Add missing type annotations for properties

cdce8p:
This one would have been my last typing PR. Unfortunately though, it requires support for different property `getter` and `setter` types, which mypy doesn't have (yet). If those attributes were only used internally, it wouldn't be a big deal to add a few type ignores. However as these are also accessed and set from outside, it's probably not helpful to add them at this time. If the mypy issue ever get's resolved, the PR could be resurrected. Until then, I'll probably close this PR in a few days.

- https://github.com/python/mypy/issues/3004

### DIFF
--- a/requirements-mypy.txt
+++ b/requirements-mypy.txt
@@ -1,3 +1,3 @@
-mypy==1.15.0
+mypy==1.16.0
 types-colorama
 wcwidth-stubs
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -771,11 +771,11 @@ class PrettyTable:
         self._xhtml = val
 
     @property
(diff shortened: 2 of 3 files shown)


cdce8p:
> Unfortunately though, it requires support for different property `getter` and `setter` types, which mypy doesn't have (yet).

There is an open PR which adds support for it. https://github.com/python/mypy/pull/18510
Will likely be included in `1.16.0`.

cdce8p:
Mypy `1.16.0` was released today. https://mypy-lang.blogspot.com/2025/05/mypy-116-released.html
Included the dependency bump here. The PR is ready to go now.

hugovk:
Thanks!

PR #344, opened 2024-12-09
Fix default styles not being reset between `set_style()` calls

MonstersInc-sudo:
Looks to fix issue as outlined in [Issue 341](https://github.com/prettytable/prettytable/issues/341)

Addition of pytest tests for testing if default has been reset after initially having another style. The case of random is tested, but frankly a redundant test as other cases cover every scenario of a reset.

Added resetting of styles in the set_style method along with removal of the default_reset in the orgmode setter.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1454,6 +1454,7 @@ class PrettyTable:
 
     def set_style(self, style: TableStyle) -> None:
         self._style = style
+        self._set_default_style()
         if style == TableStyle.DEFAULT:
             self._set_default_style()
         elif style == TableStyle.MSWORD_FRIENDLY:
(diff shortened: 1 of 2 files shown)


src/prettytable/prettytable.py
```diff
@@ -1454,6 +1454,7 @@ def _get_options(self, kwargs: Mapping[str, Any]) -> OptionsType:
 
     def set_style(self, style: TableStyle) -> None:
         self._style = style
+        self._set_default_style()
```
hugovk: Little nit, so when we read the `if` statement, the `style` var is still in our head :)

```suggestion
        self._set_default_style()
        self._style = style
```

src/prettytable/prettytable.py
```diff
@@ -1454,6 +1454,7 @@ def _get_options(self, kwargs: Mapping[str, Any]) -> OptionsType:
...
         if style == TableStyle.DEFAULT:
             self._set_default_style()
         elif style == TableStyle.MSWORD_FRIENDLY:
```
hugovk: We can remove this, because we've just called `self._set_default_style`:

```suggestion
        if style == TableStyle.MSWORD_FRIENDLY:
```

hugovk:
Thank you!

Issue #272, opened 2024-01-15
argument max_width doesn't work


aisviper:
I try this code

```python
x = PrettyTable(max_width=10)
x.field_names = ["City name", "Area", "Population", "Annual Rainfall"]
x.add_row(["Adelaide", 1295, 1158259, 600.5])
x.add_row(["Brisbane", 5905, 185759411111111111111111111, 1146.4])
x.add_row(["Darwin", 112, 120900, 1714.7])
x.add_row(["Hobart", 1357, 205556, 619.5])
x.add_row(["Sydney", 2058, 4336374, 1214.8])
x.add_row(["Melbourne", 1566, 3806092, 646.9])
x.add_row(["Perth", 5386, 1554769, 869.4])

print(x)
```
and get:

```
+-----------+------+-----------------------------+-----------------+
| City name | Area |          Population         | Annual Rainfall |
+-----------+------+-----------------------------+-----------------+
|  Adelaide | 1295 |           1158259           |      600.5      |
|  Brisbane | 5905 | 185759411111111111111111111 |      1146.4     |
|   Darwin  | 112  |            120900           |      1714.7     |
|   Hobart  | 1357 |            205556           |      619.5      |
|   Sydney  | 2058 |           4336374           |      1214.8     |
| Melbourne | 1566 |           3806092           |      646.9      |
|   Perth   | 5386 |           1554769           |      869.4      |
+-----------+------+-----------------------------+-----------------+
```

if i change code to:

```python
x = PrettyTable(["City name", "Area", "Population", "Annual Rainfall"], max_width=10)
x.add_row(["Adelaide", 1295, 1158259, 600.5])
x.add_row(["Brisbane", 5905, 185759411111111111111111111, 1146.4])
x.add_row(["Darwin", 112, 120900, 1714.7])
x.add_row(["Hobart", 1357, 205556, 619.5])
x.add_row(["Sydney", 2058, 4336374, 1214.8])
x.add_row(["Melbourne", 1566, 3806092, 646.9])
x.add_row(["Perth", 5386, 1554769, 869.4])

print(x)
```
then the argument works:
```
+-----------+------+------------+-----------------+
| City name | Area | Population | Annual Rainfall |
+-----------+------+------------+-----------------+
|  Adelaide | 1295 |  1158259   |      600.5      |
|  Brisbane | 5905 | 1857594111 |      1146.4     |
|           |      | 1111111111 |                 |
|           |      |  1111111   |                 |
[...]

PR #345, opened 2025-01-01
Fix `add_autoindex` alignment for HTML

### LINKED ISSUE #199: Align Issue with add_autoindex

stuertz:
Fixes #199 using changes from  #272

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1646,8 +1646,8 @@ class PrettyTable:
         Arguments:
         fieldname - name of the field to contain the new column of data"""
         self._field_names.insert(0, fieldname)
-        self._align[fieldname] = self.align
-        self._valign[fieldname] = self.valign
+        self._align[fieldname] = self._kwargs["align"] or "c"
+        self._valign[fieldname] = self._kwargs["valign"] or "t"
         for i, row in enumerate(self._rows):
             row.insert(0, i + 1)
 
--- a/tests/test_prettytable.py
+++ b/tests/test_prettytable.py
@@ -2354,6 +2354,132 @@ class TestMaxTableWidth:
 +---+-----------------+---+-----------------+---+-----------------+""".strip()
         )
 
+    def test_table_width_on_init_wo_columns(self) -> None:
+        """See also #272"""
+        table = PrettyTable(max_width=10)
+        table.add_row(
+            [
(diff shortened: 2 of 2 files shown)


hugovk:
Thanks!

PR #346, opened 2025-01-16
Use `TableStyle` in `README.md`

yoannmos:
Update of the README to use TableStyle.MARKDOWN instead of the deprecated MARKDOWN


hugovk:
Thank you!

(Noting this deprecation was added in https://github.com/prettytable/prettytable/pull/329.)

PR #360, opened 2025-02-02
Performance: remove unnecessary deepcopies and add some lru_caches

stuertz:
which led to some improvements in test performance

from
```
  lint: OK (2.01=setup[0.03]+cmd[1.98] seconds)
  mypy: OK (2.15=setup[2.02]+cmd[0.13] seconds)
  pypy3: SKIP (0.02 seconds)
  py314: SKIP (0.01 seconds)
  py313: OK (2.92=setup[1.78]+cmd[1.15] seconds)
  py312: OK (2.88=setup[1.70]+cmd[1.18] seconds)
  py311: SKIP (0.01 seconds)
  py310: SKIP (0.01 seconds)
  py39: OK (4.96=setup[3.70]+cmd[1.26] seconds)
  congratulations :) (15.02 seconds)
```
to:
```
  lint: OK (2.06=setup[0.02]+cmd[2.04] seconds)
  mypy: OK (2.16=setup[2.04]+cmd[0.11] seconds)
  pypy3: SKIP (0.01 seconds)
  py314: SKIP (0.01 seconds)
  py313: OK (2.91=setup[1.85]+cmd[1.07] seconds)
  py312: OK (2.78=setup[1.73]+cmd[1.05] seconds)
  py311: SKIP (0.01 seconds)
  py310: SKIP (0.01 seconds)
  py39: OK (4.74=setup[3.60]+cmd[1.14] seconds)
  congratulations :) (14.72 seconds)
```
which are around 7%

in an example with 25.000 rows:
```python
import random

def add_rows(table):
    table.add_row(
        ["Adelaide", 1295, random.randint(0, 9999999), random.randint(0, 9999)]
    )
    table.add_row(
        ["Brisbane", 5905, random.randint(0, 9999999), random.randint(0, 9999)]
    )
    table.add_row(["Darwin", 112, random.randint(0, 9999999), random.randint(0, 9999)])
    table.add_row(["Hobart", 1357, random.randint(0, 9999999), random.randint(0, 9999)])
    table.add_row(["Sydney", 2058, random.randint(0, 9999999), random.randint(0, 9999)])
    table.add_row(
        ["Melbourne", 1566, random.randint(0, 9999999), random.randint(0, 9999)]
    )
    table.add_row(["Perth", 5386, random.randint(0, 9999999), random.randint(0, 9999)])

def create_table():
    table = PrettyTable(
        field_names=["City name", "Area", "Population", "Annual Rainfall"],
    )
    for i in range(5000):
        add_rows(table)
    return table

def get_table():
    t = create_table()
    return t

def main():
    t = get_table()
    s = t.get_string()

if __name__ == "__main__":
    main()

```

it improved on my machine from 1.12secs to 328msecs:

from
```
~/privat/repos/prettytable performance *1 ?2 ❯ venv/bin/python -m timeit "import test; test.main()"
1 loop, best of 5: 1.12 sec per loop
~/privat/repos/prettytable performance *1 ?2 ❯ venv/bin/python -m timeit "import test; test.main()"
1 loop, best of 5: 1.12 sec per loop
~/privat/repos/prettytable performance *1 ?2 ❯ venv/bin/python -m timeit "import test; test.main()"
1 loop, best of 5: 1.1 sec per loop
```
to
```
~/privat/repos/prettytable performance !1 ?2 ❯ venv/bin/python -m timeit "import test; test.main()"
1 loop, best of 5: 328 msec per loop
~/privat/repos/prettytable performance !1 ?2 ❯ venv/bin/python -m timeit "import test; test.main()"
1 loop, best of 5: 326 msec per loop
~/privat/repos/prettytable performance !1 ?2 ❯ venv/bin/python -m timeit "import test; test.main()"
1 loop, best of 5: 328 msec per loop
```

Those improvements are > 300%

I can't find any reason for deepcopies here. The results  of the internal `_get_rows()` function are not modified anywhere in the code. The `copy()` and the `PrettyTable()[start:end]`  functionality is not touched.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -33,7 +33,7 @@
 
 from __future__ import annotations
 
-import functools
+from functools import lru_cache
 import io
 import re
 import warnings
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -38,6 +38,7 @@ import re
 import warnings
 from collections.abc import Callable, Iterable, Mapping, Sequence
 from enum import IntEnum
+from functools import lru_cache
 from html.parser import HTMLParser
 from typing import TYPE_CHECKING, Any, Final, Literal, TypedDict, cast
 
@@ -148,7 +149,7 @@ class OptionsType(TypedDict):
 _re = re.compile(r"\033\[[0-9;]*m|\033\(B")
 
 
-@functools.lru_cache
+@lru_cache
 def _get_size(text: str) -> tuple[int, int]:
     lines = text.split("\n")
     height = len(lines)
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2705,7 +2705,7 @@ class PrettyTable:
 ##############################
 
 
-@functools.lru_cache
+@lru_cache
 def _str_block_width(val: str) -> int:
     import wcwidth  # type: ignore[import-untyped]
 
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -33,7 +33,6 @@
 
 from __future__ import annotations
 
-from functools import lru_cache
 import io
 import re
 import warnings
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -33,6 +33,7 @@
 
 from __future__ import annotations
 
+import functools
 import io
 import re
 import warnings
@@ -148,6 +149,7 @@ class OptionsType(TypedDict):
 _re = re.compile(r"\033\[[0-9;]*m|\033\(B")
 
 
+@functools.lru_cache
 def _get_size(text: str) -> tuple[int, int]:
     lines = text.split("\n")
     height = len(lines)
@@ -1861,12 +1863,11 @@ class PrettyTable:
         Arguments:
 
         options - dictionary of option settings."""
-        import copy
 
         if options["oldsortslice"]:
-            rows = copy.deepcopy(self._rows[options["start"] : options["end"]])
+            rows = self._rows[options["start"] : options["end"]]
         else:
-            rows = copy.deepcopy(self._rows)
+            rows = self._rows
 
         # Sort
         if options["sortby"]:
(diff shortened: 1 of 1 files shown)


src/prettytable/prettytable.py
```diff
@@ -33,6 +33,7 @@
 
 from __future__ import annotations
 
+import functools
```
hugovk: Let's use this form:
```suggestion
from functools import lru_cache
```

src/prettytable/prettytable.py
```diff
@@ -2707,6 +2706,7 @@ def _get_formatted_latex_string(self, options: OptionsType) -> str:
 ##############################
 
 
+@functools.lru_cache
```
```suggestion
@lru_cache
```

hugovk: Do you think using `@cache` (which has no `maxsize`) would be too much?

hugovk (commented):
Thanks!

> in an example with 25.000 rows:

This example has 7 x 5,000 = 35,000. That's an extreme example but a good way to multiply up to demonstrate the improvements.

Also good on my machine:

```console
❯ python -m timeit "import test; test.main()" # main
1 loop, best of 5: 889 msec per loop

❯ python -m timeit "import test; test.main()" # performance
1 loop, best of 5: 433 msec per loop
```

Another method, including the overhead of the Python interpreter:

```console
❯ hyperfine --warmup 3 --prepare "git checkout main" "python3 test.py # main" --prepare "git checkout performance" "python3 test.py # performance"
Benchmark 1: python3 test.py # main
  Time (mean ± σ):     933.4 ms ±  11.6 ms    [User: 912.3 ms, System: 18.5 ms]
  Range (min … max):   920.8 ms … 959.8 ms    10 runs

Benchmark 2: python3 test.py # performance
  Time (mean ± σ):     468.6 ms ±   6.5 ms    [User: 455.4 ms, System: 12.2 ms]
  Range (min … max):   459.7 ms … 483.8 ms    10 runs

Summary
  python3 test.py # performance ran
    1.99 ± 0.04 times faster than python3 test.py # main
```


> I can't find any reason for deepcopies here.

I checked git blame and got back to when the repo was imported from Google Code. I wasn't motivated to spelunk the [old Subversion repo](https://code.google.com/archive/p/prettytable/source/default/source?page=1) :)

src/prettytable/prettytable.py
stuertz: I don't think its worth it. In very large tables and long running processes you may introduce some kind of potential large "mem leak".

hugovk:
Thanks! 🚀

PR #370, opened 2025-02-16
Use `_str_block_width` to calculate title width

### LINKED ISSUE #232: Use `_str_block_width` to calculate title width
https://github.com/jazzband/prettytable/blob/db4a1e36269792b25c29be76f8ce48969b653c5e/src/prettytable/prettytable.py#L1572

Since title width is currently computed using `len`, zero-width characters in the title contribute to table width. `_str_block_width` removes some escape codes before calculating width and returns a more appropriate value. I propose replace `len(options["title"])` with `_str_block_width(options["title"])`.

stuertz:
Fixes #232.

fixes width computation of titles.


hugovk:
Thank you!

PR #378, opened 2025-02-27
Fix `IndexError` for `add_rows()` with an empty list

### LINKED ISSUE #377: IndexError: list out of range
### What did you do?
add_rows with empty list

system continues to function

### What actually happened?
`IndexError: list index out of range`

The problem is due to https://github.com/prettytable/prettytable/pull/369/files#diff-80c2caa7cabc20f01b4362a53781e3347ab3387630dc845071f4027139f69f85R1655, where `rows[-1]` raises exception if `rows` is empty. I think it's a breaking change.

* OS: 
* Python: 
* PrettyTable: 3.15.0

are
with minimal dependencies.

```python
# code goes here
```

hugovk:
Fixes https://github.com/prettytable/prettytable/issues/377.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1653,7 +1653,9 @@ class PrettyTable:
         # On rows[:-1]: Value of type "Iterable[list[Any]]" is not indexable
         for row in rows[:-1]:  # type: ignore[index]
             self.add_row(row)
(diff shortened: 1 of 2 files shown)


miluchen (approved):
Thanks for the quick fix. Looks good 👍

Alt Text

hugovk:
And released in 3.15.1 🚀

PR #395, opened 2025-07-29
Simplify empty dictionary/`None` check

hugovk:
Follow on from https://github.com/prettytable/prettytable/pull/286.


hugovk:
Lint failure is https://github.com/tox-dev/pre-commit-uv/issues/70, passes locally and in pre-commit.ci.

PR #397, opened 2025-08-09
feat: Add custom value formatter

### LINKED ISSUE #266: Negative Integer and Float in Red and positive in Green
Wondering if it is possible for the table to be content aware, so integer and floats are automatically printed in certain color for certain datatypes.

to get started, string in blue color, postiive values in green and negative values in red.

### LINKED ISSUE #391: Dynamic word wrapping
I would like to have a word wrapping, which works not on a coloumn basis with a fixed height, like described here (#130), but a one that will wrap text, for a coloumns that I deciced should have a word wrapping.

So instead of setting max_width, I would like to set something like:

```python
table = PrettyTable()
table.max_width = 500
table.add_coloumn('Foo')
table.add_coloumn('Bar', wrap=True)
```

So `'Foo'` coloumn will stay as is, at any circumstances, but the `'Bar'` wrap itself to a new lines.

### Context

I am developing a tool, that should print a table to a console, but I do not want to change width of certain coloumns, but some of them could have  new lines. Like, I never want to wrap a path or wrap a name, but I am open to wrap the description.

So I set size of tty in `table.max_width` and tell that I am fine with `'Bar'` to be wrapped if needed

----
Maybe such thing exists already, but I am not really aware.

stuertz:
With this feature I wanted to define the formatting (colorization)  of values in a single function, which  gets called by prettytable. Previous mechanisms allowed formatting only on columns

Fixes: #266
Fixes: #391

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,6 +597,33 @@ print(table.get_string(border=False))
 print(table)
 ```
 
+### Changing the appearance of your values
+
+You can modify the appearance of your values by using colors. E.g. you want positive
+values green and negative values red. The standard formatting has been already processed
+before calling this function, e.g. float values are already formatted with given
+precision. Because of this the colorization function gets two parameters, the original
+value (with its type) and the preprocessed formatted value.
+
+```python
+import colorama
+
+def colorUpYourLife(value, representation):
+    if isinstance(value,  (int, float)):
+        if value >= 0:
+            return f"{colorama.Fore.GREEN}{representation}{colorama.Fore.RESET}"
+        else:
+            return f"{colorama.Fore.RED}{representation}{colorama.Fore.RESET}"
+    if value is not None:
+        return f"{colorama.Fore.BLUE}{representation}{colorama.Fore.RESET}"
+    return representation
+
+table = PrettyTable(custom_value_format = colorUpYourLife)
+table.title = "Here be Table caption"
+table.add_row(["bla", 24, None])
+table.add_row(["blu", -1, None])
+```
+
 ### Changing the appearance of your table - with _colors_!
 
 PrettyTable has the functionality of printing your table with ANSI color codes. This
--- a/README.md
+++ b/README.md
@@ -602,13 +602,14 @@ print(table)
 You can modify the appearance of your values by using colors. E.g. you want positive
 values green and negative values red. The standard formatting has been already processed
 before calling this function, e.g. float values are already formatted with given
-precision. Because of this the colorization function gets two parameters, the original
-value (with its type) and the preprocessed formatted value.
+precision. Because of this the colorization function gets three parameters, the field
+name, the original value (with its type) and the preprocessed formatted value and must
+return the new representation.
 
 ```python
 import colorama
 
-def colorUpYourLife(value, representation):
+def colorUpYourLife(field, value, representation):
     if isinstance(value,  (int, float)):
         if value >= 0:
             return f"{colorama.Fore.GREEN}{representation}{colorama.Fore.RESET}"
--- a/README.md
+++ b/README.md
@@ -599,18 +599,19 @@ print(table)
 
 ### Changing the appearance of your values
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -597,6 +597,34 @@ print(table.get_string(border=False))
...
 ```
 
+### Changing the appearance of your values
+
+You can modify the appearance of your values by using colors. E.g. you want positive
+values green and negative values red. The standard formatting has been already processed
+before calling this function, e.g. float values are already formatted with given
+precision. Because of this the colorization function gets three parameters, the field
+name, the original value (with its type) and the preprocessed formatted value and must
+return the new representation.
```
```suggestion
You can modify the appearance of your values by using colors. For example, you want 
positive values green and negative values red. The standard formatting has been already
processed before calling this function, so for instance, float values are already
formatted to the required precision. Because of this, the colorization function gets
three parameters, the field name, the original value (with its type) and the
preprocessed formatted value, and must return the new representation.
```

README.md
```diff
@@ -597,6 +597,34 @@ print(table.get_string(border=False))
...
+
+```python
+import colorama
+
+def colorUpYourLife(field, value, representation):
+    if isinstance(value,  (int, float)):
```
```suggestion
import colorama
from prettytable import PrettyTable

def color_up_your_life(field, value, representation):
    if isinstance(value, (int, float)):
```

README.md
```diff
@@ -597,6 +597,34 @@ print(table.get_string(border=False))
...
+        if value >= 0:
+            return f"{colorama.Fore.GREEN}{representation}{colorama.Fore.RESET}"
+        else:
+            return f"{colorama.Fore.RED}{representation}{colorama.Fore.RESET}"
+    if value is not None:
+        return f"{colorama.Fore.BLUE}{representation}{colorama.Fore.RESET}"
+    return representation
+
+table = PrettyTable(custom_value_format = colorUpYourLife)
+table.title = "Here be Table caption"
```
```suggestion
table = PrettyTable(custom_value_format = color_up_your_life)
```

README.md
```diff
@@ -597,6 +597,34 @@ print(table.get_string(border=False))
 print(table)
 ```
 
+### Changing the appearance of your values
+
+You can modify the appearance of your values by using colors. E.g. you want positive
+values green and negative values red. The standard formatting has been already processed
+before calling this function, e.g. float values are already formatted with given
+precision. Because of this the colorization function gets three parameters, the field
```
hugovk: Why do we need to pass the field name? None of the examples in the README or tests use it.

stuertz: I think in some cases, you may want to modify the value not only because of the type or its value. The only identifier available is the name of the column. I came to that conclusion after I added only value based tests.

I could extend the tests or add an example to use the column name. What would you prefer? Or are you in doubt with passing the field name at all?

hugovk: Okay, that sounds reasonable.

It'd be good to add the field to an example and a test. Perhaps an ID column wouldn't need colouring?

stuertz:
should wait for #399

README.md
stuertz: Hey,
I'm in doubt about this addition at all.

1. There is no need for the `field_name`, because we could assign a function per column
2. there is already `custom_format` which does nearly the same. The only difference is the previous formatting is not reflected.
3. Even `custom_format`does not need the column at all, for the same reason.
4. Merging this into custom_format seems not possible without breaking
5. The linked issues could also be fixed with `custom_format`IMHO.

The only thing we could adopt from this would be the missing / bad documention of custom_format and its possibilities, which then should be a new Issue/MR

WDYT?

hugovk:
#399 is merged, is this ready for review?

README.md
hugovk: (Posting this after my [comment](https://github.com/prettytable/prettytable/pull/397#issuecomment-3533625680) below :)

Yeah, seems like we could skip this, and the docs in https://github.com/prettytable/prettytable/pull/405 will help.

PR #401, opened 2025-08-28
In `add_rows()`, check length of array with `len()`

### LINKED ISSUE #400: `add_rows` fails with NumPy array
### What did you do?

```python
import numpy as np
t = PrettyTable()
arr = np.array([[1, 2, 3], [11, 12, 13]])
t.add_rows(arr)
```

```pycon
>>> t.add_rows(arr)
>>> t
+---------+---------+---------+
| Field 1 | Field 2 | Field 3 |
+---------+---------+---------+
|    1    |    2    |    3    |
|    11   |    12   |    13   |
+---------+---------+---------+
```

### What actually happened?

```pytb
>>> t.add_rows(arr)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "$HOME/.local/lib/python3.10/site-packages/prettytable/prettytable.py", line 1655, in add_rows
    if rows:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

* OS: Ubuntu 22
* Python: 3.10
* PrettyTable: 3.16.0

### Cause

The problem is here:
https://github.com/prettytable/prettytable/blob/ec1373124194abbeea692cf91f7b94561a40b9a1/src/prettytable/prettytable.py#L1667-L1671

The intent is to check if the array is non-empty, but it's not allowed to evaluate an `ndarray` in boolean context.

frankkusters:
Fixes #400.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -1667,7 +1667,7 @@ class PrettyTable:
         for row in rows[:-1]:
             self.add_row(row)
 
-        if rows:
+        if len(rows) > 0:
             self.add_row(rows[-1], divider=divider)
 
     def add_row(self, row: RowType, *, divider: bool = False) -> None:


src/prettytable/prettytable.py
```diff
@@ -1667,7 +1667,7 @@ def add_rows(self, rows: Sequence[RowType], *, divider: bool = False) -> None:
         for row in rows[:-1]:
             self.add_row(row)
 
-        if rows:
+        if len(rows) > 0:
```
hugovk, line 1670: What do you think about this, for some symmetry with the first loop?
```suggestion
        for row in rows[-1:]:
```

hugovk (commented):
Thanks! We don't explicitly support NumPy and also don't test it, but this fix is okay as we're really checking if there's anything left in rows.

src/prettytable/prettytable.py
frankkusters: `rows[-1:]` actually means "an array which contains the last element of `rows`". So `for row in rows[-1:]` means "loop over the last element of `rows`". If you do that, then this whole block should change to:

```python
for row in rows[:-1]:
    self.add_row(row)

for row in rows[-1:]:
    self.add_row(row, divider=divider)
```

Even though it's symmetrical, to me it is weird to loop over "an array with the last element, which might be empty if the array was empty". It makes the whole if-condition _implicit_, which I don't think is a good thing.

hugovk:
Thanks!

PR #412, opened 2025-11-14
Replace pre-commit with prek

hugovk:
https://hugovk.dev/blog/2025/ready-prek-go/

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: "3.x"
-      - uses: tox-dev/action-pre-commit-uv@v1
+      - uses: j178/prek-action@v1
 
   mypy:
     runs-on: ubuntu-latest
--- a/tox.ini
+++ b/tox.ini
@@ -25,11 +25,9 @@ commands =
 [testenv:lint]
 skip_install = true
 deps =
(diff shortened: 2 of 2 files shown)


ulgens:
Hey 👋 Not directly related to the changes, but I have a question: Is it needed to have prek running for both tox in also lint workplace? I feel like having it in tox is redundant because in CI, it will be already running with the lint workflow, and locally, with git hooks.

hugovk:
I'm not quite sure I follow the question, but let me try :)

Locally, I do use the Git hooks, but I recognise not everyone likes pre-commit, and this is why it's also checked on the CI. It's also possible to make changes through the GitHub UI, so you of course don't have pre-commit running, and there's also the dependency update PRs.

It's also in tox, because I think it's a good idea to have tox environments to cover being able to test and lint. So a contributor doesn't need to start trying to figure out and install whatever tools are needed for each thing. It'll take care of it for you. Plus do it all in a venv.

I don't usually run plain `tox`, but usually the thing I want, like `tox -e py314`, or `tox -e py310`, or `tox -e mypy` and so on.

Does that help?

ulgens:
> because I think it's a good idea to have tox environments to cover being able to test and lint. So a contributor doesn't need to start trying to figure out and install whatever tools are needed for each thing.

That's what I was looking for. Thank you ✨

PR #416, opened 2025-11-30
Remove unnecessary layers from lint action

ulgens:
Removes unnecessary `actions/setup-python` layers from lint workflow

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,9 +16,6 @@ jobs:
       - uses: actions/checkout@v5
         with:
           persist-credentials: false
-      - uses: actions/setup-python@v6
-        with:
-          python-version: "3.x"
       - uses: j178/prek-action@v1
 
   mypy:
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -28,10 +25,7 @@ jobs:
       - uses: actions/checkout@v5
         with:
           persist-credentials: false
-      - uses: actions/setup-python@v6
-        with:
-          python-version: "3.x"
-      - name: Install uv
```
hugovk, line 34: Hmm, the docs say:

> Using actions/setup-python can be faster, because GitHub caches the Python versions alongside the runner.

https://github.com/astral-sh/setup-uv#do-i-still-need-actionssetup-python-alongside-setup-uv

ulgens: Hey, sorry for the late reply. I did some testing and it seems there isn't any significant gain with using multiple layers: https://github.com/ulgens/ruffen-docs/pull/11

Then I checked the setup-uv's issue tracker to understand the history and found [this comment](https://github.com/astral-sh/setup-uv/issues/197#issuecomment-2552291883):

> ...and might even save ~1s. But for most usecases you don't have to use it.

and it seems this part didn't make it into the docs.

I created an issue in setup-uv repo to clarify the situation: https://github.com/astral-sh/setup-uv/issues/724 , but my understanding is that the setup-python layer is not necessary, unless you care about a potential ~1s improvement.

PR #419, opened 2025-12-31
Allow different types for Cursor than only sqlite3.Cursor

### LINKED ISSUE #415: Cursor type check failure
### What did you do?
Run `mypy` to check typing via `poetry run mypy`.

Type checking to succeed

### What actually happened?
Got the following error:
```
error: Argument 1 to "from_db_cursor" has incompatible type "psycopg.cursor.Cursor[tuple[Any, ...]]"; expected "sqlite3.Cursor"  [arg-type]
```

I am currently using `psycopg` as my Postgres adapter, which implements the Python DB API 2.0 specifications. Documentation [here](https://github.com/prettytable/prettytable/blob/main/src/prettytable/prettytable.py#L46) claims that we can use a library as long as it conforms with the Python DB API, which `psychopg` does. However, the type assumes that we use `sqlite3`:
https://github.com/prettytable/prettytable/blob/main/src/prettytable/prettytable.py#L46

Ideally, we can make the type more generic.

* OS: macOS Tahoe 26.1
* Python: 3.13.7
* PrettyTable: 3.16.0

are
with minimal dependencies.

```python
def example(cursor: psycopg.Cursor[tuple[Any, ...]]):
    tbl = prettytable.from_db_cursor(cursor)
```

stuertz:
Allow `Any` for cursor, as there is no common base class between cursors of psycopng and sqlite3 other than `object`

Fixes #415

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -43,7 +43,6 @@ from typing import Any, Literal, TypedDict, cast
 TYPE_CHECKING = False
 if TYPE_CHECKING:
     from collections.abc import Callable, Mapping, Sequence
-    from sqlite3 import Cursor
     from typing import Final, TypeAlias
 
     from _typeshed import SupportsRichComparison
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks!

PR #425, opened 2026-01-04
Overwrite int/float formatter with `custom_formatter`

### LINKED ISSUE #247: custom_format does not override float_format
Not sure this is the right tag for this issue, but here goes:
When using `custom_format` of a column that's already been formatted with `float_format`, the `custom_format` setting is ignored.

When using `custom_format` on various float data, it works as intended. In my case, I would like to set a global format with `float_format`, then override a single column with `custom_format`

I've tried to loop over a dictionary that contains my various format settings, but I'm not well-versed in lambda functions.

stuertz:
Fixes #247 

Remove existing float / int / None formatters for column if a custom_formatter is applied.

So either a custom_formatter or a type specific formatter can be set.

Used the same pattern in  custom_formatter setter as with float and int formatters

It's not an ideal solution, because only one formatter should be installed for a column at any time. But that would require a larger rewrite.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -97,11 +97,34 @@ HeaderStyleType: TypeAlias = Literal["cap", "title", "upper", "lower"] | None
 
 
 class ObservableDict(dict[str, Any]):
+    """A dictionary that notifies a callback when items are set or changed.
+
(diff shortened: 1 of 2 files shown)


hugovk:
Thanks, this seems like a good improvement.

PR #426, opened 2026-01-04
Fix columns used in tests and check real result

### LINKED ISSUE #424: Improve Tests
### What did you do?
The test in 
```python
    def test_set_for_one_column(self, city_data: PrettyTable) -> None:
        city_data.align["Rainfall"] = "l"
        city_data.max_width["Name"] = 10
        city_data.int_format["Population"] = "4"
        city_data.float_format["Area"] = "2.2"
        assert city_data.get_string() == city_data[:].get_string()
```

is fundamentally broken. 

* Most of the column names do not exist
* The setting of the options do not have an effect, only for `align` if applied to the correct column `Annual Rainfall` 

There are several things to do:
* fix column names in test
* expect the changes to have an effect 

* OS: MacOS
* Python:  3.14
* PrettyTable: main

stuertz:
fixes #424

### DIFF
--- a/tests/test_prettytable.py
+++ b/tests/test_prettytable.py
@@ -415,11 +415,48 @@ class TestOptionAttribute:
         assert city_data.get_string() == city_data[:].get_string()
 
     def test_set_for_one_column(self, city_data: PrettyTable) -> None:
-        city_data.align["Rainfall"] = "l"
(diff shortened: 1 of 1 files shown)


hugovk:
Thanks!

PR #429, opened 2026-01-10
Validate column-specific options when dict attrs are modified

### LINKED ISSUE #202: Alignment not validated when set for specific field

stuertz:
# Call validators for column specific options if dict attrs where modified

E.g. 
```python 
table["Field 1].align = "rice"
```

should raise an Exception, because of an invalid value

* fixes #202 
* Something similar, only for align/valign was tried in #207

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -130,6 +130,7 @@ class ObservableDict(dict[str, Any]):
             self.callback(key, old_value, value)
         super().__setitem__(key, value)
 
+
 class OptionsType(TypedDict):
     title: str | None
     start: int
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -126,10 +126,9 @@ class ObservableDict(dict[str, Any]):
             value: The new value to set
         """
         old_value = self.get(key)
-        super().__setitem__(key, value)
         if self.callback is not None and old_value != value:
             self.callback(key, old_value, value)
-
+        super().__setitem__(key, value)
 
 class OptionsType(TypedDict):
     title: str | None
@@ -385,6 +384,19 @@ class PrettyTable:
 
         self._custom_format: dict[str, Callable[[str, Any], str]] = ObservableDict()
         self._custom_format.callback = self._custom_format_callback
+
+        self._align: dict[str, str | None] = ObservableDict()
+        self._align[BASE_ALIGN_VALUE] = "c"
+        self._align.callback = self._align_callback
+
+        self._valign: dict[str, str | None] = ObservableDict()
+        self._valign.callback = self._valign_callback
+
+        self._max_width: dict[str, str | None] = ObservableDict()
+        self._max_width.callback = self._max_width_callback
+
+        self._min_width: dict[str, str | None] = ObservableDict()
+        self._min_width.callback = self._min_width_callback
 
         self._kwargs = {}
         if field_names:
@@ -877,6 +889,20 @@ class PrettyTable:
         else:
             self.valign = "t"
 
+    def _align_callback(self, field_name, old_value, new_value):
+        """Callback to call validators if dict attrs are modified.
+
+        This callback is triggered when a field is modified from align dict and
+        calls the validator for the new val.
+
+        Arguments:
+            field_name: Name of the field being removed
+            old_value: Previous value (unused)
+            new_value: New value (unused)
+
+        """
+        self._validate_align(new_value)
+
     @property
     def align(self) -> dict[str, AlignType]:
         """Controls alignment of fields
@@ -888,7 +914,6 @@ class PrettyTable:
     @align.setter
     def align(self, val: AlignType | dict[str, AlignType] | None) -> None:
         if isinstance(val, str):
-            self._validate_align(val)
             if not self._field_names:
                 self._align = {BASE_ALIGN_VALUE: val}
             else:
@@ -896,7 +921,6 @@ class PrettyTable:
                     self._align[field] = val
         elif isinstance(val, dict) and val:
             for field, fval in val.items():
-                self._validate_align(fval)
                 self._align[field] = fval
         else:
             if not self._field_names:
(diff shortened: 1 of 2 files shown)


src/prettytable/prettytable.py
```diff
@@ -877,6 +886,20 @@ def field_names(self, val: Sequence[Any]) -> None:
         else:
             self.valign = "t"
 
+    def _align_callback(self, field_name, old_value, new_value):
+        """Callback to call validators if dict attrs are modified.
+
+        This callback is triggered when a field is modified from align dict and
+        calls the validator for the new val.
```
```suggestion
        calls the validator for the new value.
```

src/prettytable/prettytable.py
```diff
@@ -905,6 +926,20 @@ def align(self, val: AlignType | dict[str, AlignType] | None) -> None:
                 for field in self._field_names:
                     self._align[field] = "c"
 
+    def _valign_callback(self, field_name, old_value, new_value):
+        """Callback to call validators if dict attrs are modified.
+
+        This callback is triggered when a field is modified from valign dict
+        and calls the validator for the new val.
```
```suggestion
        and calls the validator for the new value.
```

src/prettytable/prettytable.py
```diff
@@ -916,19 +951,31 @@ def valign(self) -> dict[str, VAlignType]:
...
                 self._valign[field] = fval
         else:
             for field in self._field_names:
                 self._valign[field] = "t"
 
+    def _max_width_callback(self, field_name, old_value, new_value):
+        """Callback to call validators if dict attrs are modified.
+
+        This callback is triggered when a field is modified from max_width dict
+        and calls the validator for the new val.
```
```suggestion
        and calls the validator for the new value.
```

src/prettytable/prettytable.py
```diff
@@ -940,15 +987,27 @@ def max_width(self) -> dict[str, int]:
...
                 self._max_width[field] = fval
         else:
-            self._max_width = {}
+            self._max_width.clear()
+
+    def _min_width_callback(self, field_name, old_value, new_value):
+        """Callback to call validators if dict attrs are modified.
+
+        This callback is triggered when a field is modified from min_width dict
+        and calls the validator for the new val.
```
```suggestion
        and calls the validator for the new value.
```

src/prettytable/prettytable.py
```diff
@@ -386,6 +382,19 @@ def __init__(self, field_names: Sequence[str] | None = None, **kwargs) -> None:
...
+        self._align[BASE_ALIGN_VALUE] = "c"
+        self._align.callback = self._align_callback
+
+        self._valign: dict[str, str | None] = ObservableDict()
+        self._valign.callback = self._valign_callback
+
+        self._max_width: dict[str, str | None] = ObservableDict()
+        self._max_width.callback = self._max_width_callback
+
+        self._min_width: dict[str, str | None] = ObservableDict()
```
hugovk: ?
```suggestion
        self._max_width: dict[str, int | None] = ObservableDict()
        self._max_width.callback = self._max_width_callback

        self._min_width: dict[str, int | None] = ObservableDict()
```

src/prettytable/prettytable.py
```diff
@@ -877,6 +886,20 @@ def field_names(self, val: Sequence[Any]) -> None:
...
             self.valign = "t"
 
+    def _align_callback(self, field_name, old_value, new_value):
+        """Callback to call validators if dict attrs are modified.
+
+        This callback is triggered when a field is modified from align dict and
+        calls the validator for the new val.
+
+        Arguments:
+            field_name: Name of the field being removed
```
```suggestion
            field_name: Name of the field being modified
```

src/prettytable/prettytable.py
```diff
@@ -905,6 +926,20 @@ def align(self, val: AlignType | dict[str, AlignType] | None) -> None:
...
                     self._align[field] = "c"
 
+    def _valign_callback(self, field_name, old_value, new_value):
+        """Callback to call validators if dict attrs are modified.
+
+        This callback is triggered when a field is modified from valign dict
+        and calls the validator for the new val.
+
+        Arguments:
+            field_name: Name of the field being removed
```
```suggestion
            field_name: Name of the field being modified
```

src/prettytable/prettytable.py
```diff
@@ -916,19 +951,31 @@ def valign(self) -> dict[str, VAlignType]:
...
                 self._valign[field] = "t"
 
+    def _max_width_callback(self, field_name, old_value, new_value):
+        """Callback to call validators if dict attrs are modified.
+
+        This callback is triggered when a field is modified from max_width dict
+        and calls the validator for the new val.
+
+        Arguments:
+            field_name: Name of the field being removed
```
```suggestion
            field_name: Name of the field being modified
```

src/prettytable/prettytable.py
```diff
@@ -940,15 +987,27 @@ def max_width(self) -> dict[str, int]:
...
+            self._max_width.clear()
+
+    def _min_width_callback(self, field_name, old_value, new_value):
+        """Callback to call validators if dict attrs are modified.
+
+        This callback is triggered when a field is modified from min_width dict
+        and calls the validator for the new val.
+
+        Arguments:
+            field_name: Name of the field being removed
```
```suggestion
            field_name: Name of the field being modified
```

tests/test_prettytable.py
```diff
@@ -393,6 +393,32 @@ def test_aligned_mediawiki(
...
++-----------+------+------------+-----------------+""".strip()
+        )
+
+    def test_aligned_one_column_invalid(self, city_data: PrettyTable) -> None:
+        with pytest.raises(ValueError):
+            city_data.align["Population"] = "rice"  # type: ignore[assignment]
+
+    def test_aligned_one_column_invalid_dict(self, city_data: PrettyTable) -> None:
+        with pytest.raises(ValueError):
+            city_data.align = {"Population": "rice"}  # type: ignore[dict-item]
```
hugovk: Maybe make it more explicit?

```suggestion
            city_data.align["Population"] = "unexpected"  # type: ignore[assignment]

    def test_aligned_one_column_invalid_dict(self, city_data: PrettyTable) -> None:
        with pytest.raises(ValueError):
            city_data.align = {"Population": "unexpected"}  # type: ignore[dict-item]
```

Shall we also test `valign`, `max_width`, and `min_width`?

hugovk (commented):
Thanks, looks good.

tests/test_prettytable.py
stuertz: I thought it would only blow up the tests. We don't test all options everywhere.

stuertz:
Thx for the review

tests/test_prettytable.py
hugovk: Yep, fair enough 👍

hugovk:
Thank you!

PR #432, opened 2026-01-17
Use `extend-select` instead of `select`

DimitriPapadopoulos:
Build on the Ruff default rule set. See:
* [Rule Selection](https://docs.astral.sh/ruff/tutorial/#rule-selection)
  > By default, Ruff enables Flake8's `F` rules, along with a subset of the `E` rules, omitting any stylistic rules that overlap with the use of a formatter, like `ruff format` or [Black](https://github.com/psf/black).
* [`extend-select`](https://docs.astral.sh/ruff/settings/#lint_extend-select)
  > A list of rule codes or prefixes to enable, in addition to those specified by [`select`](https://docs.astral.sh/ruff/settings/#lint_select).
* [`select`](https://docs.astral.sh/ruff/settings/#lint_select)
  > **Default value:** `["E4", "E7", "E9", "F"]`


hugovk:
Thanks!

PR #434, opened 2026-01-18
Start applying ruff/flake8-bugbear rules (B)

DimitriPapadopoulos:
I have not applied these rules:
* [unused-loop-control-variable (B007)](https://docs.astral.sh/ruff/rules/unused-loop-control-variable/#unused-loop-control-variable-b007)
* [raise-without-from-inside-except (B904)](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/#raise-without-from-inside-except-b904)
* [zip-without-explicit-strict (B905)](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/#zip-without-explicit-strict-b905)

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -537,7 +537,6 @@ class PrettyTable:
         new.field_names = self.field_names
         for attr in self._options:
             setattr(new, "_" + attr, getattr(self, "_" + attr))
-        new._align = self._align
         if isinstance(index, slice):
             for row in self._rows[index]:
                 new.add_row(row)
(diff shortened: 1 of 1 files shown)


src/prettytable/prettytable.py
```diff
@@ -537,7 +537,7 @@ def __getitem__(self, index: int | slice) -> PrettyTable:
         new.field_names = self.field_names
         for attr in self._options:
             setattr(new, "_" + attr, getattr(self, "_" + attr))
-        setattr(new, "_align", getattr(self, "_align"))
+        new._align = self._align
```
stuertz: Intersting change. Why is it neccessary at all? `align` is also an option as all the other options.
I think this line  is obsolete.

DimitriPapadopoulos: Ah, indeed. This line originates in 32d4cd3 back in 2013, I wasn't able to track it further, I think this used to be a Subversion repository at the time.

PR #436, opened 2026-01-18
Start applying Pylint rules (PLR)

DimitriPapadopoulos:
I have not applied these Refactor rules:
* all PLR09* rules
* [magic-value-comparison (PLR2004)](https://docs.astral.sh/ruff/rules/magic-value-comparison/#magic-value-comparison-plr2004)

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -897,11 +897,11 @@ class PrettyTable:
             for field, fval in val.items():
                 self._validate_align(fval)
                 self._align[field] = fval
-        elif not self._field_names:
-            self._align = {BASE_ALIGN_VALUE: "c"}
-        else:
+        elif self._field_names:
             for field in self._field_names:
                 self._align[field] = "c"
+        else:
+            self._align = {BASE_ALIGN_VALUE: "c"}
 
     @property
     def valign(self) -> dict[str, VAlignType]:
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -504,20 +504,19 @@ class PrettyTable:
             return text + excess * " "
         elif align == "r":
             return excess * " " + text
-        else:
-            if excess % 2:
-                # Uneven padding
-                # Put more space on right if text is of odd length...
-                if _str_block_width(text) % 2:
-                    return (excess // 2) * " " + text + (excess // 2 + 1) * " "
-                # and more space on left if text is of even length
-                else:
-                    return (excess // 2 + 1) * " " + text + (excess // 2) * " "
-                # Why distribute extra space this way?  To match the behaviour of
-                # the inbuilt str.center() method.
+        elif excess % 2:
+            # Uneven padding
+            # Put more space on right if text is of odd length...
+            if _str_block_width(text) % 2:
+                return (excess // 2) * " " + text + (excess // 2 + 1) * " "
+            # and more space on left if text is of even length
             else:
-                # Equal padding on either side
-                return (excess // 2) * " " + text + (excess // 2) * " "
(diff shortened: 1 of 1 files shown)


src/prettytable/prettytable.py
```diff
@@ -504,20 +504,19 @@ def _justify(self, text: str, width: int, align: AlignType) -> str:
...
-                    return (excess // 2 + 1) * " " + text + (excess // 2) * " "
-                # Why distribute extra space this way?  To match the behaviour of
-                # the inbuilt str.center() method.
+        elif excess % 2:
+            # Uneven padding
+            # Put more space on right if text is of odd length...
+            if _str_block_width(text) % 2:
+                return (excess // 2) * " " + text + (excess // 2 + 1) * " "
+            # and more space on left if text is of even length
             else:
```
stuertz, line 513: I think this else is also obsolete

DimitriPapadopoulos:
I'm sorry, I slightly changed the first commit because of an oversight, I hadn't noticed you had started looking into the PR.

src/prettytable/prettytable.py
DimitriPapadopoulos: Why is it obsolete? In any case, [`str.center`](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str) still behaves like this:
```python
>>> excess = 5
>>> 
>>> odd_length = "12345"
>>> width = excess + len(odd_length)
>>> odd_length.center(width, '*')
'**12345***'
>>> 
>>> even_length = "123456"
>>> width = excess + len(even_length)
>>> even_length.center(width, '*')
'***123456**'
>>> 
```

stuertz: I meant this else
```python
            if _str_block_width(text) % 2:
                return (excess // 2) * " " + text + (excess // 2 + 1) * " "
            else:
                return (excess // 2 + 1) * " " + text + (excess // 2) * " "
```
can be changed to
```python
            if _str_block_width(text) % 2:
                return (excess // 2) * " " + text + (excess // 2 + 1) * " "
            return (excess // 2 + 1) * " " + text + (excess // 2) * " "
```

DimitriPapadopoulos: Got it. This would be enforced by the Ruff rule [superfluous-else-return (RET505)](https://docs.astral.sh/ruff/rules/superfluous-else-return/#superfluous-else-return-ret505), which I haven't applied so far because some maintainers find the resulting less readable (lacking symmetry).

How about enforcing RET rules in a different PR?

hugovk: Yeah, I do kind of prefer that symmetry, it can be easier to read.

src/prettytable/prettytable.py
```diff
@@ -898,12 +897,11 @@ def align(self, val: AlignType | dict[str, AlignType] | None) -> None:
...
+        elif not self._field_names:
+            self._align = {BASE_ALIGN_VALUE: "c"}
         else:
-            if not self._field_names:
-                self._align = {BASE_ALIGN_VALUE: "c"}
-            else:
-                for field in self._field_names:
-                    self._align[field] = "c"
+            for field in self._field_names:
+                self._align[field] = "c"
```
hugovk: Let's flip the condition first before flattening, to remove a `not`:
```python
        elif self._field_names:
                self._align[field] = "c"
        else:
            self._align = {BASE_ALIGN_VALUE: "c"}
```

hugovk (commented):
I've run pylint on projects before and applied _some_ of the fixes, but never added it to the lint/CI before, because I felt some of the warnings weren't an improvement.

These look okay, with one suggestion.

PR #437, opened 2026-01-18
Simpler strings

DimitriPapadopoulos:
* Use f-string interpolation instead of concatenation.
* Use raw strings for literals containing `"\\"`.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -537,7 +537,7 @@ class PrettyTable:
         new = PrettyTable()
         new.field_names = self.field_names
         for attr in self._options:
-            setattr(new, "_" + attr, getattr(self, "_" + attr))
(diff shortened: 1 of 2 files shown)


src/prettytable/prettytable.py
```diff
@@ -2934,13 +2934,13 @@ def get_mediawiki_string(self, **kwargs) -> str:
             and options["attributes"]
         ):
             attr_str = " ".join(f'{k}="{v}"' for k, v in options["attributes"].items())
-            lines.append("{| " + attr_str)
+            lines.append(f"{{| {attr_str}")
```
DimitriPapadopoulos, line 2983: The `{{` might be less readable. Not sure about this one.

DimitriPapadopoulos:
Rebased.

src/prettytable/prettytable.py
hugovk: Also not sure, but let's do it :)

PR #440, opened 2026-01-25
Improve wcwidth>=0.3.5 integration and performance

jquast:
Starting [version 0.3.0](https://wcwidth.readthedocs.io/en/latest/intro.html#history), wcwidth has built-in [ljust()](https://wcwidth.readthedocs.io/en/latest/api.html#wcwidth.ljust), rjust(), center(), and a friendly[ width()](https://wcwidth.readthedocs.io/en/latest/intro.html#width) that parses ansi sequences pretty much the same way, even the hyperlinks, even the parity-odd center() justification.

To be clear, all previous and new tests also succeed with older version of wcwidth before this change, it is just moving most code responsibility to wcwidth, but I do suggest it as it has "fast path" tricks that have been benchmarked to improve performance.

One test parameter would fail with previous wcwidth integration:

        # control code (bell)
        ("abc\x07def", 6),

The ascii bell caused the previous call to wcswidth() to return -1, not a big deal.

Many new tests are provided as `.txt` files so that you may `cat tests/data/*.txt` to visually confirm them.


pyproject.toml
```diff
@@ -30,7 +30,7 @@ classifiers = [
 ]
 dynamic = [ "version" ]
 dependencies = [
-  "wcwidth",
+  "wcwidth>=0.3.5,<1",
```
hugovk: I prefer not to use upper bounds: https://iscinumpy.dev/post/bound-version-constraints/

```suggestion
  "wcwidth>=0.3.5",
```

tests/test_style.py
```diff
@@ -585,7 +594,92 @@ def test_link_and_color() -> None:
...
+    rows: list[list[str]],
+    align: Literal["l", "c", "r"] | None,
+    expected_file: str,
+) -> None:
+    table = PrettyTable(fields)
+    for row in rows:
+        table.add_row(row)
+    if align:
+        table.align[fields[0]] = align
+    with open(os.path.join(DATA_DIR, expected_file)) as fin:
```
hugovk: Test failed on Windows for 3.10-3.14:

```
E       UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 39: character maps to <undefined>
```

UTF-8 is the default in 3.15: https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-utf8-default


This should fix it:
```suggestion
    with open(os.path.join(DATA_DIR, expected_file), encoding="utf-8") as fin:
```

tests/test_style.py
```diff
@@ -585,7 +594,92 @@ def test_link_and_color() -> None:
...
+    ],
+)
+def test_table_alignment_with_emoji(
+    align: Literal["l", "c", "r"], expected_file: str
+) -> None:
+    table = PrettyTable(["Name"])
+    table.align["Name"] = align
+    table.add_row(["\U0001f468\u200d\U0001f469\u200d\U0001f467"])  # 👨‍👩‍👧
+    table.add_row(["Hi"])
+    with open(os.path.join(DATA_DIR, expected_file)) as fin:
```
hugovk:
```suggestion
    with open(os.path.join(DATA_DIR, expected_file), encoding="utf-8") as fin:
```

hugovk (commented):
Thanks! I was looking at switching to `width`, but also nice to use `ljust`, `rjust` and `center` as well :)

tests/test_style.py
```diff
@@ -585,7 +594,92 @@ def test_link_and_color() -> None:
...
+    fields: list[str],
+    rows: list[list[str]],
+    align: Literal["l", "c", "r"] | None,
+    expected_file: str,
+) -> None:
+    table = PrettyTable(fields)
+    for row in rows:
+        table.add_row(row)
+    if align:
+        table.align[fields[0]] = align
```
hugovk: This isn't run:

https://app.codecov.io/gh/prettytable/prettytable/pull/440?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=prettytable

Shall we delete it or add a test case?

hugovk:
(I put "Changelog: Changed" label, so this performance improvement will show up in release notes.)

tests/test_style.py
jquast: Deleted it, the "align" test was put into the next test case, nothing of value lost

hugovk:
Thank you!

PR #443, opened 2026-01-27
Replace `re.split` with `str.split` and remove `re` import

hugovk:
https://github.com/prettytable/prettytable/pull/440 removed some regexes, and the remaining two don't need to be regexes:

```python
[cell.strip() for cell in re.split(r"\s*!!\s*", line[1:])]
```

We call `strip()` on each `cell`, so the `\s*` handling is redundant.

Let's use `str.split` instead and remove the `re` import.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -32,7 +32,6 @@
 from __future__ import annotations
 
 import io
-import re
 from enum import IntEnum
 from functools import lru_cache
 from html.parser import HTMLParser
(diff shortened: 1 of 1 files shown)


hugovk:
`str.split` is also 2.5x faster than `re.split`:

```console
❯ python -m timeit -s 'line = "| A || B || C"; import re' '[cell.strip() for cell in re.split(r"\s*\|\|\s*", line[1:])]'
500000 loops, best of 5: 474 nsec per loop

❯ python -m timeit -s 'line = "| A || B || C"' '[cell.strip() for cell in line[1:].split("||")]'
2000000 loops, best of 5: 191 nsec per loop
```

PR #447, opened 2026-03-01
fix: respect zero-value for left/right padding width

### LINKED ISSUE #220: Removing all padding still results in padding existing.
### What did you do?

Set all padding to zero, and removed all VRules

```
============================================================
Name                              Played     Won       Chips
============================================================
xxxxxxxxxxxxxxxxxxxx                  xxasdfasdfzxcvzxcvzxcv
1                   xxxxxxxxxxxxxxxxxxxx      xx           4
============================================================
```

### What actually happened?

```
=================================================================
 Name                               Played      Won        Chips 
=================================================================
 xxxxxxxxxxxxxxxxxxxx                   xx asdfasdf zxcvzxcvzxcv 
 1                    xxxxxxxxxxxxxxxxxxxx       xx            4 
=================================================================
```

### What I was trying to actually achieve, but struggled to find the options for:

```
===============================================================
Name                               Played      Won        Chips
===============================================================
xxxxxxxxxxxxxxxxxxxx                   xx asdfasdf zxcvzxcvzxcv
1                    xxxxxxxxxxxxxxxxxxxx       xx            4
===============================================================

```

Basically, it appear impossible to create tables on PrettyTable that have zero padding, namely because space padding characters are added automatically. see: https://github.com/jazzband/prettytable/blob/master/src/prettytable/prettytable.py#L1869

even if it would be intuitive for zero padding to mean minimum padding for internal padding, it doesn't make sense for it to exist for the outside edges.

As it is, it's acting like what I would have expected if the vrule character was set to " " and the cross characters set to "="

Jo2234:
## Summary

Fix `left_padding_width=0` and `right_padding_width=0` being silently ignored.

## Problem

```python
t = PrettyTable()
t.field_names = ['Name', 'Value']
t.add_row(['foo', 'bar'])
t.padding_width = 0
t.left_padding_width = 0
t.right_padding_width = 0
print(t)
# Still shows padding!
```

## Root Cause

In `__init__` (line 453-454):

```python
self._left_padding_width = kwargs['left_padding_width'] or None
self._right_padding_width = kwargs['right_padding_width'] or None
```

Since `0` is falsy in Python, `0 or None` evaluates to `None`. When `_left_padding_width` is `None`, the code falls back to `_padding_width` (default 1), making it impossible to set true zero padding via the constructor.

## Fix

```diff
-  self._left_padding_width = kwargs['left_padding_width'] or None
-  self._right_padding_width = kwargs['right_padding_width'] or None
+  self._left_padding_width = (
+      kwargs['left_padding_width']
+      if kwargs['left_padding_width'] is not None
+      else None
+  )
+  self._right_padding_width = (
+      kwargs['right_padding_width']
+      if kwargs['right_padding_width'] is not None
+      else None
+  )
```

Fixes #220

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -450,8 +450,16 @@ class PrettyTable:
             self._padding_width = 1
         else:
             self._padding_width = kwargs["padding_width"]
-        self._left_padding_width = kwargs["left_padding_width"] or None
-        self._right_padding_width = kwargs["right_padding_width"] or None
+        self._left_padding_width = (
+            kwargs["left_padding_width"]
+            if kwargs["left_padding_width"] is not None
+            else None
+        )
+        self._right_padding_width = (
+            kwargs["right_padding_width"]
+            if kwargs["right_padding_width"] is not None
+            else None
+        )
 
         self._vertical_char = kwargs["vertical_char"] or "|"
         self._horizontal_char = kwargs["horizontal_char"] or "-"


hugovk:
Your example shows no change.

With prettytable==3.17.0:
```pycon
>>> from prettytable import PrettyTable
... t = PrettyTable()
... t.field_names = ['Name', 'Value']
... t.add_row(['foo', 'bar'])
... t.padding_width = 0
... t.left_padding_width = 0
... t.right_padding_width = 0
... print(t)
... # Still shows padding!
...
+----+-----+
|Name|Value|
+----+-----+
|foo | bar |
+----+-----+
```
With this PR:
```pycon
>>> from prettytable import PrettyTable
... t = PrettyTable()
... t.field_names = ['Name', 'Value']
... t.add_row(['foo', 'bar'])
... t.padding_width = 0
... t.left_padding_width = 0
... t.right_padding_width = 0
... print(t)
... # Still shows padding!
...
+----+-----+
|Name|Value|
+----+-----+
|foo | bar |
+----+-----+
```

PR #452, opened 2026-03-30
Fix bleeding of wrapped colours into table edges

jquast:
Problem
------------

Standard python ``textwrap.wrap()`` was used, it is not "sequence-aware".

- "bleeding" of colors into the edges of table, eg. '║'
- mismeasurements cause textwrap.wrap to wrap very early, too early
- probably also mismeasurements of emojis, regional flags, OSC hyperlinks, possible breaking on graphemes, corruption by "hard wrapp" of very long sequences, etc.

prettytable-bleeding-and-badly-wrapped

> bleeding of orange color and too-early wrapping of orange text seen near red mouse cursor

Solution
------------

use [wcwidth.wrap](https://wcwidth.readthedocs.io/en/latest/intro.html#wrap) function that handles measuring and breaking on graphemes and terminal output sequences correctly

prettytable-fixed

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2408,7 +2408,7 @@ class PrettyTable:
         return "".join(bits)
 
     def _stringify_row(self, row: list[str], options: OptionsType, hrule: str) -> str:
-        import textwrap
+        import wcwidth
 
         for index, field, value, width in zip(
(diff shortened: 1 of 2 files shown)


hugovk (approved):
wcwidth saves the day again! Thanks!

PR #459, opened 2026-04-08
Improve type coverage of `colortable`

jorenham:
This increases the global type coverage of `prettytable` from 76.38% to 83.14% (according to [typestats](https://github.com/jorenham/typestats))

If you're open to it, I'd also like to take a look at other annotations as well. For example, the `PrettyTable.__init__` kwargs are currenty untyped, so I was thinking of using a `TypedDict` to annotate those options ([PEP 692](https://peps.python.org/pep-0692/) can be used because already because of the `__future__.annotations` import ;) ).

### DIFF
--- a/src/prettytable/colortable.py
+++ b/src/prettytable/colortable.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, ClassVar, Final
+from .prettytable import PrettyTable
 
+TYPE_CHECKING = False
 if TYPE_CHECKING:
     from collections.abc import Sequence
-
-from .prettytable import PrettyTable
+    from typing import ClassVar, Final
 
 try:
     from colorama import init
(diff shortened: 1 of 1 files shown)


src/prettytable/colortable.py
```diff
@@ -1,5 +1,10 @@
 from __future__ import annotations
 
+from typing import TYPE_CHECKING, ClassVar, Final
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
 from .prettytable import PrettyTable
```
hugovk: I know we can't avoid the import entirely in this project, but I find it easier to follow the samepatter in all projects:

```suggestion
from .prettytable import PrettyTable

TYPE_CHECKING = False
if TYPE_CHECKING:
    from collections.abc import Sequence
    from typing import ClassVar, Final
```

hugovk:
Thanks!

PR #465, opened 2026-05-24
Drop stale `align`/`valign` keys when `field_names` are renamed

SAY-5:
Renaming a table via the `field_names` setter never removed the previous field names from `self._align` and `self._valign`.

The cleanup loops guarded the `pop` with `if old_name not in self._align`, so the branch only ran when the key was already absent and the old keys were left behind. After a rename the alignment dicts kept growing with dead keys.

This switches the guard to `if old_name not in val` (the new names) and uses `pop(..., None)` so overlapping names are preserved and absent keys do not raise.

Added two regression tests in `TestAlignment` covering a full rename and a partial-overlap rename.


hugovk:
Thanks!

PR #468, opened 2026-06-16
Expand tabs in cell values so columns stay aligned

### LINKED ISSUE #113: Tabulators break formatting
I am trying to print a diff between different commits in Git. The string contains a tabulator character that seems to interfere with the centering.
Example string from diff:
```diff
diff --git program.c program.c
index 05ced0c..804cd7d 100644
--- partdiff.c
+++ partdiff.c
@@ -377,6 +377,7 @@ main (int argc, char** argv)
        struct calculation_arguments arguments;
        struct calculation_results results;

+       printf("Some random text!");
        askParams(&options, argc, argv);

        initVariables(&arguments, &results, &options);
```

Actually spelled out the "\t":
```diff
diff --git partdiff.c partdiff.c
index 05ced0c..804cd7d 100644
--- partdiff.c
+++ partdiff.c
@@ -377,6 +377,7 @@ main (int argc, char** argv)
 \tstruct calculation_arguments arguments;
 \tstruct calculation_results results;

+\tprintf("Some random text!");
 \taskParams(&options, argc, argv);

 \tinitVariables(&arguments, &results, &options);
```
Printing with a PrettyTable breaks the layout. If you replace the "\t" with any character it works fine,
```
|   4   | 2021-05-13 17:41:21.512372 | BUILD |    make   |       0       |       f0252dd       |         diff --git partdiff.c partdiff.c         |     |
|       |                            |       |           |               |                     |          index 05ced0c..804cd7d 100644           |     |
|       |                            |       |           |               |                     |                  --- partdiff.c                  |     |
|       |                            |       |           |               |                     |                  +++ partdiff.c                  |     |
|       |                            |       |           |               |                     | @@ -377,6 +377,7 @@ main (int argc, char** argv) |     |
|       |                            |       |           |               |                     |                                struct calculation_arguments arguments;                          |     |
|       |                            |       |           |               |                     |                                struct calculation_results results;                          |     |
|       |                            |       |           |               |                     |                                                  |     |
|       |                            |       |           |               |                     |                         +      printf("Some random text!");                          |     |
|       |                            |       |           |               |                     |                                askParams(&options, argc, argv);                          |     |
|       |                            |       |           |               |                     |                                                  |     |
|       |                            |       |           |               |                     |                                initVariables(&arguments, &results, &options);                          |     |
|       |                            |       |           |               |                     |                                                  |     |
+-------+----------------------------+-------+-----------+---------------+---------------------+--------------------------------------------------+-----+

```

gaoflow:
Fixes #113.

A cell that contains a tab character is rendered misaligned, because the width PrettyTable measures for it does not match the width it displays at:

```python

t = PrettyTable()
t.field_names = ["code", "note"]
t.add_row(["if x:\n\treturn 1", "tab-indented"])
print(t.get_string())
```

On current `main` this prints (the tab line is short, so the column does not line up):

```
+------------------+--------------+
|       code       |     note     |
+------------------+--------------+
|      if x:       | tab-indented |
| 	return 1 |              |
+------------------+--------------+
```

A `\t` is one column wide to `wcwidth`, so `_str_block_width` under-measures the line, but a terminal advances the tab to the next tab stop on display. The measured width and the rendered width disagree, so the padding is wrong.

(For reference: this is still reproducible after #440's wcwidth work, so the issue is not yet resolved.)

## Fix

Expand tabs once in `_format_value`, which is the single point that produces the cell string used for *both* width measurement and rendering (`_format_rows` feeds `_compute_widths` and `_stringify_row`). With the cell expanded up front, both paths see the same string and agree, so the column lines up:

```
+------------------+--------------+
|       code       |     note     |
+------------------+--------------+
|      if x:       | tab-indented |
|         return 1 |              |
+------------------+--------------+
```

`str.expandtabs()` uses the standard tab size of 8, matching the terminal tab-stop convention. Cells without tabs are unchanged.

## Tests

Added `TestWidth::test_tab_expansion_aligns_columns`, which asserts a tab-containing cell renders with no raw tab and with every line the same display width. It is RED without the fix and GREEN with it; the full suite passes (336 passed).

Disclosure: I prepared this fix with AI assistance under my direction; I reviewed and verified the change and the test myself.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2019,9 +2019,18 @@ class PrettyTable:
             return (f"%{self._float_format[field]}f") % value
 
         formatter = self._custom_format.get(field, (lambda f, v: str(v)))
-        # Expand tabs so the width we measure matches what a terminal renders;
-        # a raw "\t" is one character wide to wcwidth but advances to the next
-        # tab stop on display, which otherwise misaligns the column (issue #113).
+        # Prettytable is unaware of a terminal's tabstops, and it does not know at
+        # what specific location of the screen it will be displayed, so it also cannot
+        # calculate tabstop positions or width: A '\t' character is variable-width,
+        # depending on the location of the screen it is displayed.
+        #
+        # Although wcwidth library functions like width() do measure tab control
+        # character as a width of 8, it would require prettytable to display the left
+        # margin of the table's contents to begin "at the tabstop" for the table
+        # contents and dividers to line up correctly.
+        #
+        # PrettyTable is "screen unaware", so it is best to alter tabstops to a fixed
+        # width, to allow same-width display anywhere on the screen.
         return formatter(field, value).expandtabs()
 
     def _compute_table_width(self, options) -> int:
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2019,7 +2019,10 @@ class PrettyTable:
             return (f"%{self._float_format[field]}f") % value
 
         formatter = self._custom_format.get(field, (lambda f, v: str(v)))
-        return formatter(field, value)
+        # Expand tabs so the width we measure matches what a terminal renders;
+        # a raw "\t" is one character wide to wcwidth but advances to the next
+        # tab stop on display, which otherwise misaligns the column (issue #113).
+        return formatter(field, value).expandtabs()
 
     def _compute_table_width(self, options) -> int:
         if options["vrules"] == VRuleStyle.FRAME:
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2019,13 +2019,13 @@ class PrettyTable:
             return (f"%{self._float_format[field]}f") % value
 
         formatter = self._custom_format.get(field, (lambda f, v: str(v)))
-        # Prettytable is unaware of a terminal's tabstops, and it does not know at
+        # PrettyTable is unaware of a terminal's tabstops, and it does not know at
         # what specific location of the screen it will be displayed, so it also cannot
         # calculate tabstop positions or width: A '\t' character is variable-width,
         # depending on the location of the screen it is displayed.
         #
         # Although wcwidth library functions like width() do measure tab control
-        # character as a width of 8, it would require prettytable to display the left
+        # character as a width of 8, it would require PrettyTable to display the left
         # margin of the table's contents to begin "at the tabstop" for the table
         # contents and dividers to line up correctly.
         #
--- a/tests/test_prettytable.py
+++ b/tests/test_prettytable.py
@@ -1611,10 +1611,7 @@ class TestWidth:
 """.strip()
 
     def test_tab_expansion_aligns_columns(self) -> None:
-        # Issue #113: a tab in a cell is one character wide to wcwidth but
-        # advances to the next tab stop when rendered, so it must be expanded
-        # for the measured width to match the rendered width. Otherwise the
-        # tab line is shorter than the rest and the column misaligns.
+        """Ensure expandtabs() is used in PrettyTable._format_value()."""
         table = PrettyTable(["code", "note"])
         table.add_row(["if x:\n\treturn 1", "tab-indented"])
         result = table.get_string()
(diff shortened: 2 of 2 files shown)


src/prettytable/prettytable.py
```diff
@@ -2019,7 +2019,10 @@ def _format_value(self, field: str, value: Any) -> str:
             return (f"%{self._float_format[field]}f") % value
 
         formatter = self._custom_format.get(field, (lambda f, v: str(v)))
-        return formatter(field, value)
+        # Expand tabs so the width we measure matches what a terminal renders;
+        # a raw "\t" is one character wide to wcwidth but advances to the next
+        # tab stop on display, which otherwise misaligns the column (issue #113).
```
jquast: [wcwidth.width()](https://wcwidth.readthedocs.io/en/latest/api.html#wcwidth.width) and its higher level functions like wrap() or ljust() used by prettytable measures tabs as a width of 8, not 1. the [wcwidth.wcwidth()](https://wcwidth.readthedocs.io/en/latest/api.html#wcwidth.wcwidth) function would return **negative 1**, not 1 as suggested. 

The exact size of ``\t`` depends on the location of ``\t`` within the text for wcwidth functions that support it, just the same as [str.expandtabs](https://docs.python.org/3/library/stdtypes.html#str.expandtabs) does: it as though the first column of the given text will be displayed at a tabstop (or the far left margin).

It might surprise you to see a tab expanded to only 1 space:

```python
>>> 'abcdefg\t'
'abcdefg\t'
>>> 'abcdefg\t'.expandtabs()
'abcdefg '
```

```suggestion
        # Prettytable is unaware of a terminal's tabstops, and it does not know at
        # what specific location of the screen it will be displayed, so it also cannot
        # calculate tabstop positions or width: A '\t' character is variable-width,
        # depending on the location of the screen it is displayed.
        #
        # Although wcwidth library functions like width() do measure tab control
        # character as a width of 8, it would require prettytable to display the left
        # margin of the table's contents to begin "at the tabstop" for the table
        # contents and dividers to line up correctly.
        #
        # PrettyTable is "screen unaware", so it is best to alter tabstops to a fixed
        # width, to allow same-width display anywhere on the screen.
```

tests/test_prettytable.py
```diff
@@ -1610,6 +1610,18 @@ def test_reset(self) -> None:
 +---------+---------+
 """.strip()
 
+    def test_tab_expansion_aligns_columns(self) -> None:
+        # Issue #113: a tab in a cell is one character wide to wcwidth but
+        # advances to the next tab stop when rendered, so it must be expanded
+        # for the measured width to match the rendered width. Otherwise the
+        # tab line is shorter than the rest and the column misaligns.
```
jquast:
```suggestion
        """Ensure expandtabs() is used in PrettyTable._format_value()."""
```

gaoflow:
Thanks @jquast, you're right about the wcwidth measurements (`wcwidth('\t')` is -1 and `width()` treats it as 8). I applied both suggestions in 42eb9ed: the screen-unaware rationale on the comment and the docstring on the test. Happy to adjust further if you'd like.

hugovk:
Thanks both!

src/prettytable/prettytable.py
```diff
@@ -2019,7 +2019,19 @@ def _format_value(self, field: str, value: Any) -> str:
             return (f"%{self._float_format[field]}f") % value
 
         formatter = self._custom_format.get(field, (lambda f, v: str(v)))
-        return formatter(field, value)
+        # Prettytable is unaware of a terminal's tabstops, and it does not know at
```
```suggestion
        # PrettyTable is unaware of a terminal's tabstops, and it does not know at
```

src/prettytable/prettytable.py
```diff
@@ -2019,7 +2019,19 @@ def _format_value(self, field: str, value: Any) -> str:
...
+        # what specific location of the screen it will be displayed, so it also cannot
+        # calculate tabstop positions or width: A '\t' character is variable-width,
+        # depending on the location of the screen it is displayed.
+        #
+        # Although wcwidth library functions like width() do measure tab control
+        # character as a width of 8, it would require prettytable to display the left
```
```suggestion
        # character as a width of 8, it would require PrettyTable to display the left
```

hugovk (approved):

PR #473, opened 2026-06-27
fix: remove extra spaces when vrules=NONE with padding=0

### LINKED ISSUE #220: Removing all padding still results in padding existing.

C1-BA-B1-F3:
## Description

This PR fixes #220 - when setting all padding to zero and removing all VRules, there were still extra spaces in the output.

### Root Cause

When , the code was replacing vertical rule characters with spaces instead of removing them entirely. This meant:
- A leading space was added at the start of each row
- Spaces were added between columns
- These spaces could not be removed by setting 

### Fix

Modified three methods in :
1.  - Skip adding separator spaces when 
2.  - Same fix for header rows
3.  - Skip adding separator characters in horizontal rules when 

### Tests

- Updated existing tests to match the new correct behavior
- Added 3 new regression tests for the fix

Fixes #220

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2330,10 +2330,14 @@ class PrettyTable:
         lpad, rpad = self._get_padding_widths(options)
         if options["vrules"] in (VRuleStyle.ALL, VRuleStyle.FRAME):
             bits = [options[f"{where}left_junction_char"]]  # type: ignore[literal-required]
+        elif options["vrules"] == VRuleStyle.NONE:
+            bits = []
         else:
             bits = [options["horizontal_char"]]
         # For tables with no data or fieldnames
         if not self._field_names:
+            if options["vrules"] == VRuleStyle.NONE:
+                return ""
             bits.append(options[f"{where}right_junction_char"])  # type: ignore[literal-required]
             return "".join(bits)
         for field, width in zip(self._field_names, self._widths):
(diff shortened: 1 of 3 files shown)


C1-BA-B1-F3:
Checking pre-commit failure...

PR #475, opened 2026-06-30
fix: handle rows with fewer cells than header in from_html()

### LINKED ISSUE #474: from_html() raises ValueError on a row with fewer cells than the header (off-by-one in padding)
### Description

### Reproduction

```python

```

### Expected vs actual

### Root cause

uttam12331:
from_html() raises ValueError when a table row has fewer cells than the header row, due to an off-by-one in the padding loop. This fixes the loop bounds and adds a regression test.

Closes #474

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -3200,7 +3200,7 @@ def _make_table_handler():
             for row in self.rows:
                 if len(row[0]) < self.max_row_width:
                     appends = self.max_row_width - len(row[0])
-                    for i in range(1, appends):
+                    for i in range(appends):
                         row[0].append("-")
 
                 if row[1]:
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -22,6 +22,18 @@ class TestHtmlConstructor:
         with pytest.raises(ValueError):
             from_html_one(html_string)
 
+    def test_html_ragged_row(self) -> None:
+        """from_html() must not raise when a data row has fewer cells than the header."""
+        html = "<table><tr><th>a</th><th>b</th></tr><tr><td>1</td></tr></table>"
+        tables = from_html(html)
+        assert len(tables) == 1
+        table = tables[0]
+        assert table.field_names == ["a", "b"]
+        rows = table._rows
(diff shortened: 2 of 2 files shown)


hugovk:
Close/re-open to trigger GitHub Actions. Possibly didn't fire the first time due to GH anti-AI-spam measures?

tests/test_html.py
```diff
@@ -22,6 +22,18 @@ def test_html_one_fail_on_many(self, city_data: PrettyTable) -> None:
...
+        """from_html() must not raise when a data row has fewer cells than the header."""
+        html = "<table><tr><th>a</th><th>b</th></tr><tr><td>1</td></tr></table>"
+        tables = from_html(html)
+        assert len(tables) == 1
+        table = tables[0]
+        assert table.field_names == ["a", "b"]
+        rows = table._rows
+        assert len(rows) == 1
+        assert rows[0][0] == "1"
+        assert rows[0][1] == "-"
```
hugovk, line 35: Shorted docstring, use Arrange/Act/Assert, assert against `table.rows` rather than `table._rows`:

```suggestion
        """from_html() must not raise when a row has fewer cells than the header."""
        # Arrange
        html = "<table><tr><th>a</th><th>b</th></tr><tr><td>1</td></tr></table>"

        # Act
        tables = from_html(html)

        # Assert
        assert len(tables) == 1
        table = tables[0]
        assert table.field_names == ["a", "b"]
        assert table.rows == [["1", "-"]]
```

src/prettytable/prettytable.py
```diff
@@ -3200,7 +3200,7 @@ def generate_table(self, rows: list[tuple[list[str], bool]]) -> PrettyTable:
             for row in self.rows:
                 if len(row[0]) < self.max_row_width:
                     appends = self.max_row_width - len(row[0])
-                    for i in range(1, appends):
+                    for i in range(appends):
                         row[0].append("-")
```
hugovk, line 3204: Slight performance improvement for longer rows.
```suggestion
                    row[0].extend(["-"] * appends)
```

PR #476, opened 2026-06-30
Truncate wide-character headers by display width

mahirhir:
When `use_header_width=False` (or `max_table_width` shrinks a column), a field name wider than its column is truncated to fit. That truncation uses a character slice (`fieldname[:width]`), so wide characters such as CJK overflow the column and break alignment, because each CJK character occupies two display columns but `[:width]` keeps `width` characters.

```python

t = PrettyTable(["日本語ヘッダー", "col2"], use_header_width=False)
t.add_row(["abcd", "yy"])
print(t)
```

Before:

```
+------+----+
| 日本語ヘ | co |
+------+----+
| abcd | yy |
+------+----+
```

The header row is wider than the border and the data row. After the fix the table stays aligned:

```
+------+----+
| 日本 | co |
+------+----+
| abcd | yy |
+------+----+
```

The truncation guard already measures display width via `_str_block_width` (`wcwidth.width`), and `_stringify_row` already wraps cell content by display width with `wcwidth.wrap`, so the header path was the odd one out. The fix swaps the character slice for `wcwidth.clip(fieldname, 0, width)`, which truncates by display column and is already a dependency. ASCII headers are unaffected (`wcwidth.clip` matches `[:width]` for narrow text).

Added a regression test next to `test_table_max_width_wo_header_width`. The full suite passes locally.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2388,6 +2388,8 @@ class PrettyTable:
         return "\n".join(lines)
 
     def _stringify_header(self, options: OptionsType) -> str:
+        import wcwidth
+
         bits: list[str] = []
         lpad, rpad = self._get_padding_widths(options)
         if options["border"]:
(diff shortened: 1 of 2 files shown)


mahirhir:
Closing this to clean up my open PRs. It's a small change and not really worth taking up your review time. Sorry for the noise.

PR #479, opened 2026-07-04
pad short rows correctly in from_html

### LINKED ISSUE #474: from_html() raises ValueError on a row with fewer cells than the header (off-by-one in padding)

CodingSelim:
Fixes #474.

`from_html` is supposed to normalise ragged tables, but the short-row padding loop ran `range(1, appends)`, so it always padded one cell short and padded nothing at all when a row was missing exactly one cell. That made it raise `ValueError` on a row with fewer cells than the header.

```python
from_html("<table><tr><th>a</th><th>b</th></tr><tr><td>1</td></tr></table>")  # used to raise
```

Changed the loop to `range(appends)` and added a test.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -3200,7 +3200,7 @@ def _make_table_handler():
             for row in self.rows:
                 if len(row[0]) < self.max_row_width:
                     appends = self.max_row_width - len(row[0])
-                    for i in range(1, appends):
+                    for _ in range(appends):
                         row[0].append("-")
 
                 if row[1]:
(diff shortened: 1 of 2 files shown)


hugovk:
There's already https://github.com/prettytable/prettytable/pull/475 open to fix this. Why is this PR better than that one?

CodingSelim:
not better really, its the same one line fix, i just missed 475 when i searched. only real difference is my test asserts through the public .rows instead of ._rows so go with theirs and pull that bit in if you want, closing this one.

PR #483, opened 2026-07-18
fix: avoid IndexError when sizing a zero-column titled table

santhreal:
PrettyTable._compute_widths hits IndexError when get_string(title=...) with zero columns. This PR fixes the regression with a focused test covering the case.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -2118,7 +2118,7 @@ class PrettyTable:
             # What is being scaled is content so we sum column widths
             content_width = sum(widths) or 1
 
-            if content_width < min_width:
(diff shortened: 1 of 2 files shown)


hugovk:
You've opened many PRs across these repos two weeks ago, and then closed them with no comment:

https://github.com/python-humanize/humanize/pull/352
https://github.com/python-humanize/humanize/pull/353
https://github.com/prettytable/prettytable/pull/483
https://github.com/prettytable/prettytable/pull/484
https://github.com/jazzband/tablib/pull/659
https://github.com/jazzband/tablib/pull/660
https://github.com/jazzband/tablib/pull/661

Please could you explain why?

Also hundreds in other repos:

image

PR #489, opened 2026-07-31
Fix colspan DoS in from_html()

lilu5458:
## Summary

Fixes an unbounded `colspan` memory-exhaustion DoS (CWE-409) in the `from_html()` HTML parser.

### Unbounded `colspan` causes memory exhaustion DoS (CWE-409)

The `_TableHandler` parser parsed the `colspan` attribute via `int(value)` without any bounds checking. A malicious HTML input with a very large `colspan` value (e.g. `colspan="100000000"`) caused the parser to attempt allocating ~100M list entries in:

```python
for _ in range(1, self.colspan):
    self.last_row.append("")
```

This leads to memory exhaustion and process hang/kill.

**Fix:** Validate the integer value (raising `ValueError` on invalid input) and clamp `colspan` to the HTML spec valid range `[1, 1000]`.

## Testing

- All 342 existing tests pass
- Added 4 new unit tests covering: colspan clamping, invalid colspan, negative colspan, and normal colspan

---

Note: As requested in the review, this PR has been reduced to only the colspan DoS fix. The off-by-one row-padding issue (#474) will be handled separately in #475.

### DIFF
--- a/src/prettytable/prettytable.py
+++ b/src/prettytable/prettytable.py
@@ -3165,8 +3165,15 @@ def _make_table_handler():
             if tag == "th":
                 self.is_last_row_header = True
             for key, value in attrs:
-                if key == "colspan":
-                    self.colspan = int(value)  # type: ignore[arg-type]
+                if key == "colspan" and value:
+                    try:
+                        span = int(value)
+                    except ValueError:
+                        msg = f"Invalid colspan value: {value!r}"
+                        raise ValueError(msg)
+                    # Clamp to HTML spec valid range [1, 1000] to prevent
+                    # memory exhaustion from malicious large values (CWE-409)
+                    self.colspan = max(1, min(span, 1000))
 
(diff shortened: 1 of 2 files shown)


hugovk:
This is the third PR for https://github.com/prettytable/prettytable/issues/474. Why is this PR better than the others?

hugovk:
Please reduce to just fix number 1, we'll do number 2 in https://github.com/prettytable/prettytable/pull/475.

lilu5458:
Thanks @hugovk! I've reduced this PR to only fix number 1 (the colspan DoS / CWE-409 fix) and dropped the row-padding change (#474) so it can be handled separately in #475.

The branch is now a single commit that only touches the `colspan` handling in `handle_starttag` (validate + clamp to `[1, 1000]`), with the `generate_table()` padding loop left unchanged. The row-padding test was removed; the 4 colspan tests remain. All 342 tests pass.

PR #493, opened 2026-08-18
Add first_col_is_header option to use the first column as row headers in HTML

### LINKED ISSUE #38: Add option to make first column of table the header
The only option available currently is to make the first row of the table the header. I have a use case where I prefer the first column of the table to be a header, and then the second column be the data of my table.

Are you open to adding a flag to get_html_string which swaps the header to being vertical?

Happy to contribute a PR

ChrisJr404:
Closes #38.

Adds a `first_col_is_header` option (default `False`, so existing output is unchanged) to `get_html_string`. When enabled, PrettyTable renders the first cell of every data row as a `<th scope="row">` element instead of `<td>`, and marks the column headers with `scope="col"`. This is the accessible "row header" markup from the [W3C tables tutorial](https://www.w3.org/WAI/tutorials/tables/two-headers/), which #38 asked for.

It works with both the plain and `format=True` HTML renderers, can be set via the constructor, the `table.first_col_is_header` property, or a per-call keyword argument, and follows the first *rendered* column when `fields` reorders/filters columns.

Example:

```python
table = PrettyTable()
table.field_names = ["City", "Area", "Population"]
table.add_row(["Adelaide", 1295, 1158259])
print(table.get_html_string(first_col_is_header=True))
```

```html
<table>
    <thead>
        <tr>
            <th scope="col">City</th>
            <th scope="col">Area</th>
            <th scope="col">Population</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <th scope="row">Adelaide</th>
            <td>1295</td>
            <td>1158259</td>
        </tr>
    </tbody>
</table>
```

### Notes

- Default behaviour is unchanged: with `first_col_is_header=False` no `scope` attributes are emitted and the body stays all `<td>` (covered by a regression test).
- Column headers stay in `<thead>`; only the `scope` attributes are added in this mode, so the markup remains valid for tables that have both column headers and per-row headers.

### Tests

Added five tests in `tests/test_html.py` covering the plain renderer, the formatted renderer, `fields` interaction, the default-off invariant, and the property setter/validation. Full suite passes (`343 passed`); `black` and `mypy` are clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,6 +578,7 @@ Table-specific options are:
 | `bottom_left_junction_char`  | Single character string used to draw bottom-left line junctions. Default: `junction_char`.                                                  |
 | `bottom_right_junction_char` | Single character string used to draw bottom-right line junctions. Default: `junction_char`.                                                 |
 | `break_on_hyphens`           | Whether long lines are wrapped on hyphens. Default: `True`.                                                                                 |
+| `first_col_is_header`        | A Boolean option (must be `True` or `False`). Renders the first cell of each HTML data row as a `<th scope="row">`. Default: `False`.       |
 | `header`                     | A Boolean option (must be `True` or `False`). Controls whether the first row of the table is a header showing the names of all the fields.  |
 | `horizontal_char`            | Single character string used to draw horizontal lines. Default: `-`.                                                                        |
 | `hrules`                     | Controls printing of horizontal rules after rows. Allowed values: `FRAME`, `HEADER`, `ALL`, `NONE`.                                         |
--- a/README.md
+++ b/README.md
@@ -844,6 +844,39 @@ when sending output to HTML. This can be disabled by setting the `escape_header`
 print(table.get_html_string(escape_header=False, escape_data=False))
 ```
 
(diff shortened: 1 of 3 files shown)


stuertz:
* Please describe the new option in the `README.md`in the `Table-specific options` section
* https://github.com/prettytable/prettytable/issues/38#issuecomment-1247365992 asks for json support, any thoughts on this?

ChrisJr404:
Thanks for the review!

I've added `first_col_is_header` to the **Table-specific options** table in `README.md` (alphabetically, right after `break_on_hyphens`), in the same style as the other Boolean entries and noting the `False` default. There's also a short worked example a bit further down in the HTML section.

On JSON output: I think that's a genuinely useful feature, but it feels like its own thing rather than something to fold into this HTML-focused change — it needs decisions about shape (list of row objects vs. columns, how to handle `None`/custom formatters, field ordering) that are orthogonal to row-header markup. I'd suggest keeping it as a separate issue/PR so it can get proper design discussion. Happy to take a look at it separately if that'd help.

